PR #1340, opened 2022-03-18
Feature/gitpod

collymore:
Added a Gitpod file to allow developing with GraphCommerce in Gitpod


collymore:
Here's a preview of how it works if you're not familiar with gitpod 
https://www.youtube.com/watch?v=aTIzFMj9p0Y

.gitpod.yml
```diff
@@ -0,0 +1,27 @@
+tasks:
+  - before: nvm install 14
+    init: |
+      mkdir my-project
+      cp -R examples/magento-graphcms/. my-project
+      cd my-project
```
paales, line 6: If we just want to spin up an example, it is also possible to just do:

```suggestion
      cd examples/magento-graphcms
```

.gitpod.yml
```diff
@@ -0,0 +1,27 @@
...
+      jq 'del(.scripts) | with_entries( if .key | contains("scripts_local") then .key |= sub("scripts_local";"scripts") else . end)' package.json > _.json && mv _.json package.json
+      cp -R .env.example .env
+      rm CHANGELOG.md && touch CHANGELOG.md
```
paales, line 9: Make it inline with the most recent 
```suggestion
      rm CHANGELOG.md
```

paales (changes requested):
Very cool @collymore! I haven't used gitpod per sé, but it would be very nice to add support for it (as well as all the other cloud editors like stackblitz / codesandbox).

I'm running into an issue when I'm using it, that is because we have a `NEXT_PUBLIC_GRAPHQL_ENDPOINT` variable, let me see if we can do something about that.

Schermafbeelding 2022-03-21 om 16 23 56

.gitpod.yml
```diff
@@ -0,0 +1,27 @@
...
+      rm -rf node_modules && rm -rf .next
+      yarn
+      yarn codegen
+    command: cd /workspace/graphcommerce/my-project && npm run dev
```
paales, line 13:
```suggestion
    command: cd /workspace/graphcommerce/examples/magento-graphcms && npm run dev
```

paales:
@collymore We do have the `GITPOD_WORKSPACE_URL` env variable available. And the eventual URL the instance runs on is `3000-$GITPOD_WORKSPACE_URL`. Maybe we can do something with that?

collymore:
> @collymore We do have the `GITPOD_WORKSPACE_URL` env variable available. And the eventual URL the instance runs on is `3000-$GITPOD_WORKSPACE_URL`. Maybe we can do something with that?

Yes that will always be the correct URL. Also there is  ```gp url``` available this will print the current url on the cli 
Screenshot 2022-03-21 at 17 03 34

collymore:
@paales https://github.com/collymore/graphcommerce/commit/a1bb3917900a8d8d9a9a9d0204c6b3d7f4e73f01 fixes the NEXT_PUBLIC_GRAPHQL_ENDPOINT variable issue

paales:
@collymore There are two directions we can go with this:

1. The current 'root' graphcommerce .gitpod assumes the monorepo is present and we are effectively setting up the monorepo and starting the current example.
2. There could be an `examples/magento-graphcms/.gitpod.yml`  that allows you to start the .gitpod by copying the example by replicating https://www.graphcommerce.org/docs/getting-started/create

I've now changed the .gitpod file to make it work with the monorepo.

paales (approved):

paales:
@collymore Merged!

PR #1343, opened 2022-03-21
Chore/canonical

ErwinOtten:


examples/magento-graphcms/pages/[...url].tsx
```diff
@@ -77,6 +77,7 @@ function CategoryPage(props: Props) {
           title={page?.metaTitle ?? ''}
           metaDescription={page?.metaDescription ?? ''}
           metaRobots={page?.metaRobots.toLowerCase().split('_').flat(1) as MetaRobots[]}
+          canonical={`/${page?.url}`}
```
Comment (paales) [inline examples/magento-graphcms/pages/[...url].tsx:80]: A page doesn't always have a URL, because it can be that there is no page defined in GraphCMS, so we need to check for it.

```suggestion
          canonical={page?.url ? `/${page?.url}` : undefined}
```

You might want to use category.url_path instead since that one will always exist

paales (changes requested):

examples/magento-graphcms/.env.example
```diff
@@ -14,8 +14,8 @@ IMAGE_DOMAINS="backend.reachdigital.dev,media.graphcms.com,media.graphassets.com
 # Used to connect to the correct GraphQL endpoint and to enable proper store switching.
 NEXT_PUBLIC_LOCALE_STORES='{"en-us": "en_US", "nl": "nl_NL", "fr-be": "fr_BE", "nl-be": "nl_BE", "en-gb": "en_GB", "en-ca": "en_CA"}'
 
-# Used as a base URL to generate the sitemaps as the base URL isn't available during buildtime.
-NEXT_PUBLIC_SITE_URL="https://graphcommerce.vercel.app"
+# Used as a base URL to generate the sitemaps and canonical url. Include trailing slash.
+NEXT_PUBLIC_SITE_URL="https://graphcommerce.vercel.app/"
```
paales, line 18:
```suggestion
# Used as a base URL to generate the sitemaps and canonical url.
NEXT_PUBLIC_SITE_URL="https://graphcommerce.vercel.app/"
```

paales (approved):

PR #1375, opened 2022-04-01
Specify PSP's so its easier to check when upgrade what one should remove

FrankHarland:


docs/getting-started/create.md
```diff
@@ -152,6 +152,7 @@ the ones your Magento backend doesn't support.
 
 - Remove `"@graphcommerce/[psp]"` from package.json
 - Remove all [psp] references from `pages/checkout/payment.tsx`
+- Currently these are: `[@graphcommerce/mollie-magento-payment, @graphcommerce/magento-payment-braintree]`
```
ErwinOtten, line 155: Is addressed mentioned at line 150

paales: Wasn't too clear for me when reading it first, having the full strings available is helpful here.

ErwinOtten: @FrankHarland suggested changes:

### Remove unused PSP's

The example has Payment Service Providers integrated. Remove
the ones your Magento backend doesn't support.

- Remove Payment Service integrations from package.json: `@graphcommerce/mollie-magento-payment` and/or `@graphcommerce/magento-payment-braintree`
- Remove Payment Service references from `pages/checkout/payment.tsx`

FrankHarland (commented):
@ErwinOtten done!

PR #1386, opened 2022-04-06
feat: add disabled support to LinkOrButton component

FrankHarland:


packages/next-ui/Button/LinkOrButton.tsx
```diff
@@ -80,6 +83,12 @@ export const LinkOrButton = React.forwardRef<
             display: { xs: 'inline-flex', [breakpoint]: 'none' },
             alignItems: 'center',
           },
+          typeof disabled !== 'undefined' &&
+            disabled && {
+              opacity: 0.5,
+              filter: 'grayscale(1)',
+              pointerEvents: 'none',
```
paales, line 90: Can we perhaps use this?

Schermafbeelding 2022-04-07 om 11 41 56

paales:
```
typeof disabled !== 'undefined' &&
            disabled && {},
```

Can also be written like:

```
!!disabled && {}
```

paales (changes requested):

packages/next-ui/Button/LinkOrButton.tsx
FrankHarland: Sure!

paales (approved):

PR #1442, opened 2022-05-05
feat: added swc as compiler, migrating away from lingui/macro

paales:


paales:
Bootup dev mode babel: 27,8 + 26,5 + 17,1 = 71,4
Bootup dev mode swc: 9+9,9+3,8 = 22,7

PR #1448, opened 2022-05-06
Fix/price formatting

timhofman:


paales:
@timhofman Can you give an example what it solves, the code doesn't make it clear :)

timhofman:
@paales we've got projects that opt for 'decimal' instead of 'currency' as formatting option to, for instance, get rid of the currency symbol. When using 'decimal' and round set to 'true', prices like 39.10 become 39.1 without this patch.

Snippet from theme.ts

`Money: {
    defaultProps: {
      round: true,
      formatOptions: {
        style: 'decimal',
        maximumFractionDigits: 2,
      },
    },
  },`

PR #1451, opened 2022-05-09
SWC: Rust based compiler to replace babel

paales:
With this PR, GraphCommerce will be using the Rust based [SWC](https://swc.rs/) compiler that will replace [babel](https://babeljs.io/). With this GraphCommerce is embracing that [Rust is the future of javascript](https://leerob.io/blog/rust).

SWC was released as part of [Next.js v12](https://nextjs.org/blog/next-12#faster-builds-and-fast-refresh-with-rust-compiler) and has been stabilizing in the last few months. With the [v12.1.1 release](https://github.com/vercel/next.js/releases/tag/v12.1.1) it came with experimental support for emotion css, so it is time to switch over.

With this PR we're also doing away with any dependencies on [babel-plugin-macros](https://github.com/kentcdodds/babel-plugin-macros) which [isn't really maintained anymore](https://github.com/kentcdodds/babel-plugin-macros/issues/177). 
This means we can't use [@lingui/macro](https://lingui.js.org/ref/macro.html) anymore. This is a bit of a downside, but we found upgrading not too difficult and we migrated the complete codebase in a few hours. And we have to migrate at some point anyways.

We are however very excited about the results of the SWC compiler!

Dev timings:

`next dev` + `babel`: 71,4s seconds
`next dev` + `swc`: 22,7 seconds

On a warm start (after next dev has run once) it is even faster: 6,1 seconds

Build times:

`next build` + `babel`: 8m 12s
`next build` + `swc`: 4m 57s

Modern preset:

We're using a little trick to also force next.js/SWC into a 'modern' mode, this means we don't compile down to ES5, but down to a much more recent version:

| Route     | Old size | New size | Diff    |
| --------- | -------- | -------- | ------- |
| /         | 272kB    | 259.0kB  | -13kB🔥 |
| /\_app    | 196kB    | 190.0kB  | -6kB🔥  |
| /[...url] | 290kB    | 276.0kB  | -14kB🔥 |

The above change isn't officially supported by next.js, but it seems to be working absolutely fine for us. You can apply the [patch](https://github.com/graphcommerce-org/graphcommerce/blob/50188e378b4c77561ebc600958ea11cd114fa61a/patches/next%2B12.1.5.patch) to your own project with the excellent [patch-package](https://www.npmjs.com/package/patch-package)


docs/framework/translations.md
```diff
@@ -26,12 +26,12 @@ msgstr "cancelado"
 
 The msgid is the message being translated. In
 /node_modules/@graphcommerce/magento-cart/components/EmptyCart/EmptyCart.tsx,
-you can see a the first msgid is passed as a propped, wrapped in the `<Trans>`
+you can see a the first msgid is passed as a the id prop to the `<Trans>`
```
ErwinOtten, line 29: you can see that the first msgid is passed as a the id prop to the `<Trans>`

docs/framework/translations.md
```diff
@@ -57,20 +57,31 @@ Refresh to see your changes updated
...
 
-```ts
+```tsx
 <Typography variant='h3'>
-  <Trans>Call us now</Trans>
+  <Trans id='Call us now' />
 </Typography>
 ```
 
+If you need to have a string instead of a React component, you can use:
```
ErwinOtten, line 68: To translate a string:

docs/framework/translations.md
```diff
@@ -57,20 +57,31 @@ Refresh to see your changes updated
...
 Add Linqui to the component's imports:
 
-```ts
-import { t, Trans } from '@lingui/macro'
+```tsx
+import { Trans } from '@lingui/react'
+import { i18n } from '@lingui/core'
 ```
 
+## Automatically extracting all translations
```
ErwinOtten, line 83: Generating translation files with all translations

docs/framework/translations.md
```diff
@@ -186,9 +197,10 @@ msgstr ""
 
 5. Add your translations ins the newly created .po file. Run the app and use the
```
ErwinOtten, line 198: Add your translations in the newly created .po file. Run the app and use the store switcher to navigate to your new storeview.

PR #1456, opened 2022-05-12
feature: add a ‘.nvmrc’ file to specify the nvm version of the project 

carlocarels90:
- run `nvm use` to use the locked nvm version of the project.


paales (commented):
We should probably also add something to the documentation how to use this functionality.

In the end you are always able to add this to your local project.

PR #1460, opened 2022-05-13
feat: mega menu

ErwinOtten:


ErwinOtten:
Replaced with: https://github.com/graphcommerce-org/graphcommerce/pull/1522

PR #1466, opened 2022-05-17
Feature/checkout address selection

FrankHarland:


examples/magento-graphcms/pages/checkout/index.tsx
```diff
@@ -81,32 +91,22 @@ function ShippingPage() {
...
-                  <li>
-                    <Trans id='Leave password field empty to order as guest.' />
-                  </li>
-                </Typography>
-              </EmailForm>
-
-              <ShippingAddressForm step={2} />
+              {addresses ? (
+                <CustomerAddressForm step={2}>
+                  <ShippingAddressForm ignoreCache step={3} />
```
FrankHarland, line 96: @paales So .. this is pretty much a DirtyFix™ (i think)
We want our form to be empty when we click the 'New address' option. I tried messing around with useFormPersist etc but could not get it to work properly. 

This works just fine, but I'm too hyped about introducing a new prop for this.

PR #1478, opened 2022-05-19
feat: add postcode-service component

NickdeK:


paales:
@NickdeK Ik denk dat er iets niet helemaal goed gaat?

paales (changes requested):
It seems there are files committed that shouldn't have been. graphql-mesh/.mesh files doesn't be there?

NickdeK:
At the moment this can't be merged because of build errors that are caused by `query_postcodeServiceNL` in the `PostcodeFields` fragment. This requires an edit in the `.meshrc.yml` which we can't do within this component/package to resolve the error.

hnsr (changes requested):
Volgens mij heeft deze package ook een dependency op @graphql-mesh/json-schema nodig

paales:
Supeseded by https://github.com/graphcommerce-org/graphcommerce/pull/1779

PR #1522, opened 2022-06-17
Feature/navigation

ErwinOtten:


paales:
These docs need to be rewritten as well? https://www.graphcommerce.org/docs/getting-started/header

PR #1551, opened 2022-07-11
feat: new cart renderer

paales:


paales:
Fixed in https://github.com/graphcommerce-org/graphcommerce/pull/2003

PR #1648, opened 2022-09-16
feat: make form labels translatable in CreateProductReviewFormProps component

LaurensFranken:


paales:
@LaurensFranken Nog even een changeset maken svp, zie deze comment: https://github.com/graphcommerce-org/graphcommerce/pull/1648#issuecomment-1249304917

PR #1672, opened 2022-09-30
add translation to paymentmethod form

FrankHarland:


.changeset/lemon-gorillas-juggle.md
```diff
@@ -0,0 +1,6 @@
+---
+"@graphcommerce/magento-cart-payment-method": patch
+"@graphcommerce/magento-cart": patch
```
paales, line 3: magento-cart can be removed

PR #1728, opened 2022-10-28
animation type spring

ErwinOtten:


packages/framer-scroller/hooks/useScrollTo.ts
```diff
@@ -34,7 +34,9 @@ export function useScrollTo() {
               },
               onComplete,
               onStop: onComplete,
-              duration: duration * 1000 || distanceAnimationDuration(ref.scrollLeft, to.x),
+              type: 'spring',
+              bounce: 0,
+              duration: duration * 2000 || distanceAnimationDuration(ref.scrollLeft, to.x),
```
paales, line 39:
```suggestion
              duration: duration * 1000 || distanceAnimationDuration(ref.scrollLeft, to.x),
```

packages/framer-scroller/hooks/useScrollTo.ts
```diff
@@ -54,7 +56,9 @@ export function useScrollTo() {
               },
               onComplete,
               onStop: onComplete,
-              duration: duration * 1000 || distanceAnimationDuration(ref.scrollTop, to.y),
+              type: 'spring',
+              bounce: 0,
+              duration: duration * 2000 || distanceAnimationDuration(ref.scrollTop, to.y),
```
paales, line 61:
```suggestion
              duration: duration * 1000 || distanceAnimationDuration(ref.scrollTop, to.y),
```

paales (changes requested):

packages/framer-scroller/hooks/useScrollTo.ts
ErwinOtten: All durations have to be adjusted if we go this route. Set it to 2000 for now to test.

paales: I dont think so, the duration is almost never provided and will almost always use the distanceAnimationDuration

PR #1777, opened 2023-01-10
New form handling

paales:
Implements a new form handling hook which reduces rererenders.

Learnings: using watch or formState causes rerenders, replace all occurences of watch with useWatch and formState with useFormState

### DIFF
--- a/packages/react-hook-form/hooks/useFormApollo.tsx
+++ b/packages/react-hook-form/hooks/useFormApollo.tsx
@@ -1,0 +1,168 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import { ApolloError, MutationOptions, QueryOptions, useApolloClient } from '@apollo/client'
+import { useMotionValueValue } from '@graphcommerce/framer-utils'
+import { isMotionValue, MotionValue, useMotionValue } from 'framer-motion'
+import {
+  FieldValues,
+  FormProvider,
+  FormProviderProps,
+  SubmitErrorHandler,
+  useForm,
+  useFormContext,
+  UseFormProps,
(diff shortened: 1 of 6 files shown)


packages/react-hook-form/hooks/useFormApollo.tsx
```diff
@@ -0,0 +1,168 @@
...
+  useFormProps?: UseFormProps<V>,
+  resetOnResult?: (result: Q) => V,
+): UseFormGqlMutationReturn<Q, V>
+export function useFormApollo<Q extends Record<string, unknown>, V extends FieldValues>(
+  operation: MutationOptions<Q, V>,
+  useFormProps?: UseFormProps<V>,
+  resetOnResult?: (result: Q) => V,
+): UseFormGqlMutationReturn<Q, V>
+export function useFormApollo<Q extends Record<string, unknown>, V extends FieldValues>(
+  operation: MutationOptions<Q, V> | QueryOptions<Partial<V>, Q>,
```
paales, line 88: - How can we support switching between multiple nutations based on external query data?
- How can we support switching between multiple mutations based on the submit button's value?
- How can we support switching between multiple mutations based on the form data?
- How can we support multiple serial mutations?
- How can we support multiple optional mutations?
- How can we support parallel mutations?
- How can we chain multiple mutations?
- How can we dedupe operations when they don't need to be executed (again)?
- How can we skip operations when they don't need to be executed because they are optional?
- How can we maintain a 'progress' on the submissions?

---

```tsx
const switchOnExternal = useFormApollo(operation: external ? One : Two)
...todo
```

PR #1792, opened 2023-02-02
Fix ellipsis for long titles on mobile

bramvanderholst:


packages/next-ui/Layout/components/LayoutHeaderContent.tsx
```diff
@@ -111,7 +111,7 @@ export function LayoutHeaderContent(props: LayoutHeaderContentProps) {
             gridTemplateAreas: `"left center right"`,
             gridTemplateColumns: '1fr auto 1fr',
             alignItems: 'center',
-            // columnGap: theme.spacings.xs,
+            gap: theme.page.horizontal,
```
paales, line 114: `theme.page.horizontal` is meant for the left/right side of the page. maybe do something like `theme.spacings.xs`?

packages/next-ui/Layout/components/LayoutHeaderContent.tsx
```diff
@@ -192,7 +192,12 @@ export function LayoutHeaderContent(props: LayoutHeaderContentProps) {
             },
           })}
         >
-          <MotionDiv layout={layout}>{children}</MotionDiv>
+          <MotionDiv
+            sx={() => ({ minWidth: 0 /* required for ellipsis to work */ })}
```
paales, line 196:
```suggestion
            sx={{ minWidth: 0 }}
```

No callback needed and the comment can be derived from the commit message.

paales (changes requested):

PR #1806, opened 2023-02-17
Fix: Navigation in DOM

ErwinOtten:
Fix issue where Navigation can't be reopened after closing, dragging or navigating

Scherm­afbeelding 2023-02-17 om 17 06 18

### DIFF
--- a/packages/next-ui/Overlay/components/OverlayBase.tsx
+++ b/packages/next-ui/Overlay/components/OverlayBase.tsx
@@ -469,7 +469,7 @@ export function OverlayBase(incomingProps: LayoutOverlayBaseProps) {
             })}
           >
             <LayoutProvider scroll={scrollYOffset}>
-              {active && (typeof children === 'function' ? children() : children)}
+              {typeof children === 'function' ? children() : children}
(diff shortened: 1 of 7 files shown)


packages/next-ui/Overlay/components/OverlayBase.tsx
```diff
@@ -469,7 +469,7 @@ export function OverlayBase(incomingProps: LayoutOverlayBaseProps) {
             })}
           >
             <LayoutProvider scroll={scrollYOffset}>
-              {active && (typeof children === 'function' ? children() : children)}
+              {typeof children === 'function' ? children() : children}
```
paales, line 472: Something like this?
 
```suggestion
              {typeof children === 'function' ? (active && children()) : children}
```

paales (changes requested):

packages/next-ui/Overlay/components/OverlayBase.tsx
ErwinOtten: Fixed

paales (approved):

PR #1809, opened 2023-02-21
Added new Google Analytics events

mikekeehnen:


packages/googleanalytics/events/gtagAddToCart/GtagAddToCart.graphql
```diff
@@ -1,5 +1,11 @@
 fragment GtagAddToCart on Cart @inject(into: ["CartItemCountChanged"]) {
   items {
+    product{
+      manufacturer
```
paales, line 4: `manufacturer` isn't a system attribute, so isn't allowed to be used.

paales (changes requested):

packages/googleanalytics/events/gtagAddToCart/GtagAddToCart.graphql
mikekeehnen: Do you know a way to get the brand of a product as a system attribute?

packages/googleanalytics/plugins/GaViewItem.tsx
```diff
@@ -10,7 +10,12 @@ export const exported = '@graphcommerce/magento-product'
 function GaViewItem(props: PluginProps<ProductPageMetaFragment>) {
   const { Prev, url_key, ...rest } = props
 
-  useEffect(() => gtagViewItem(rest), [rest])
+  useEffect(
+    () => gtagViewItem(rest),
+    // We're disabling eslint rule to prevent this event from being triggerd on every rerender
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [],
+  )
```
paales, line 18: What if we navigate from a product page to another product page, that probably wont retrigger here, right?

packages/googleanalytics/package.json
```diff
@@ -12,14 +12,14 @@
...
-    "@graphcommerce/next-ui": "5.2.0-canary.17",
-    "@graphcommerce/magento-cart-items": "5.2.0-canary.17"
+    "@graphcommerce/graphql-mesh": "6.0.0-canary.24",
+    "@graphcommerce/magento-cart": "6.0.0-canary.24",
+    "@graphcommerce/magento-cart-payment-method": "6.0.0-canary.24",
+    "@graphcommerce/magento-cart-shipping-method": "6.0.0-canary.24",
+    "@graphcommerce/magento-product": "6.0.0-canary.24",
+    "@graphcommerce/next-config": "6.0.0-canary.24",
+    "@graphcommerce/next-ui": "6.0.0-canary.24",
+    "@graphcommerce/magento-cart-items": "6.0.0-canary.24"
```
paales, line 22: I think rebasing should do the trick, then this shouldn't be here.

packagesDev/graphql-codegen-relay-optimizer-plugin/dist/module/index.d.ts
```diff
@@ -0,0 +1,4 @@
+import { PluginFunction } from '@graphql-codegen/plugin-helpers';
+export interface RelayOptimizerPluginConfig {
+}
+export declare const plugin: PluginFunction<RelayOptimizerPluginConfig>;
```
paales, line 4: I think rebasing should do the trick, then this shouldn't be here.

paales (changes requested):

packages/googleanalytics/events/gtagAddToCart/GtagAddToCart.graphql
paales: Brand isn't a system attribute unfortunately :(

All values are prefixed with `item_`, maybe allow users to inject something like this that the user can add locally?

```
fragment MyAttr on CartItem @inject(into: ["CartItemCountChanged"]) {
   product {
      item_brand: manufacturere
   }
}
```

packages/googleanalytics/plugins/GaViewItem.tsx
paales: Maybe we can do something like:

```tsx
const viewItem = useMemoDeep(rest);
useEffect(() => {}, [viewItem])
```

paales (changes requested):

packages/googleanalytics/plugins/GaCartStartCheckoutLinkOrButton.tsx
```diff
@@ -10,6 +12,18 @@ export function GaCartStartCheckoutLinkOrButton(
...
+  useEffect(
+    () =>
+      gtagViewCart({
+        items: rest.items,
+        prices: rest.prices,
+      }),
+    // We're disabling eslint rule to prevent this event from being triggerd on every rerender
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [],
+  )
```
paales, line 25: Should be replaced with useMemoObject (previously useStableObject)

PR #1822, opened 2023-02-27
Product page og:tags

KMalkowski:


packages/magento-product/components/ProductPageMeta/ProductPageMeta.tsx
```diff
@@ -3,14 +3,21 @@ import { useProductLink } from '../../hooks/useProductLink'
...
+          title={meta_title ?? name ?? ''}
+          metaDescription={meta_description ?? name ?? ''}
+          canonical={productLink}
+          sku={sku}
+          name={name}
+          image={media_gallery?.[0]?.url ?? null}
+          categories={categoryNames}
+          price={price_range?.maximum_price?.regular_price?.value ?? 0}
+          salePrice={price_range?.maximum_price?.final_price?.value ?? 0}
+      />
```
paales, line 21: Might it be an idea to do something like this?

```suggestion
      <>
      <PageMeta
          title={meta_title ?? name ?? ''}
          metaDescription={meta_description ?? name ?? ''}
          canonical={productLink}
      />
      <Head>
              {sku && <>
          <meta property='type' content='product' key='og-type' />
          <meta property='product:retailer_part_no' content={sku} key='og-sku' />
        </>}
        {name && <meta property='og:title' content={name} key='og-title' />}
        {image && <meta property='og:image' content={image} key='og-image' />}
        {price && <meta property='product:price:amount' content={price.toString()} key='og-price' />}
        {salePrice && <meta property='product:sale_price:amount' content={salePrice.toString()} key='og-sale-price' />}
        {categories && categories.map((category, index) => <meta property='product:category' content={category} key={`og-category-${index}`} />)}
      </Head>
      </>
```

paales (changes requested):

packages/magento-product/components/ProductPageMeta/ProductPageMeta.tsx
KMalkowski: Good idea, done

packages/next-ui/PageMeta/PageMeta.tsx
```diff
@@ -85,19 +86,19 @@ export function useCanonical(incoming?: Canonical) {
...
-      )}
-      <meta name='robots' content={metaRobots.join(',')} key='meta-robots' />
-      {canonical && <link rel='canonical' href={canonical} key='canonical' />}
-    </Head>
+      <Head>
+        <title>{title.trim()}</title>
+        {metaDescription && (
+            <meta name='description' content={metaDescription.trim()} key='meta-description' />
+        )}
+        {canonical && <link rel='canonical' href={canonical} key='canonical' />}
```
paales, line 100: robots seems to be removed :)

Can you also make sure indenting is set to two spaces, seems to add indenting here?

packages/magento-product/components/ProductPageMeta/ProductPageMeta.tsx
```diff
@@ -3,14 +3,25 @@ import { useProductLink } from '../../hooks/useProductLink'
...
+      {sku && <>
+          <meta property='type' content='product' key='og-type' />
+          <meta property='product:retailer_part_no' content={sku} key='og-sku' />
+      </>}
+      {name && <meta property='og:title' content={name} key='og-title' />}
+      {media_gallery?.[0]?.url && <meta property='og:image' content={media_gallery[0].url} key='og-image' />}
+      {price_range?.maximum_price?.regular_price?.value && <meta property='product:price:amount' content={price_range.maximum_price.regular_price.value.toString()} key='og-price' />}
+      {price_range?.maximum_price?.final_price?.value && <meta property='product:sale_price:amount' content={price_range.maximum_price.final_price.value.toString()} key='og-sale-price' />}
+      {categories && categories.map((category) => category?.name ? <meta property='product:category' content={category.name} key={category.uid} /> : null)}
+      {productLink && <meta property='og:url' content={productLink} key='og-url' />}
```
paales, line 24: The productLink isn't a fully qualified URL, this should be the full canonical URL. Maybe we can move this next to the canonical?

packages/magento-product/components/ProductPageMeta/ProductPageMeta.tsx
```diff
@@ -3,14 +3,25 @@ import { useProductLink } from '../../hooks/useProductLink'
...
+        canonical={productLink}
+    >
+      {sku && <>
+          <meta property='type' content='product' key='og-type' />
+          <meta property='product:retailer_part_no' content={sku} key='og-sku' />
+      </>}
+      {name && <meta property='og:title' content={name} key='og-title' />}
+      {media_gallery?.[0]?.url && <meta property='og:image' content={media_gallery[0].url} key='og-image' />}
+      {price_range?.maximum_price?.regular_price?.value && <meta property='product:price:amount' content={price_range.maximum_price.regular_price.value.toString()} key='og-price' />}
+      {price_range?.maximum_price?.final_price?.value && <meta property='product:sale_price:amount' content={price_range.maximum_price.final_price.value.toString()} key='og-sale-price' />}
```
paales, line 22: Should we also add og:currency?

paales (changes requested):
Scherm­afbeelding 2023-02-28 om 10 29 35

packages/magento-product/components/ProductPageMeta/ProductPageMeta.tsx
KMalkowski: Added product:price:currency

packages/next-ui/PageMeta/PageMeta.tsx
KMalkowski: My bad, fixed :)

packages/magento-product/components/ProductPageMeta/ProductPageMeta.tsx
KMalkowski: You're right, I've converted it here so that all the og:tags would be defined in the same place, let me know if you think it's better to use the canonical that we already have in the next-ui PageMeta and set it also to the og:url there.

KMalkowski:
@paales I've removed the spread of the category object in CategoryMeta props because the types broke after I've added children to the props. Not sure if it's the right approach

packages/magento-product/components/ProductPageMeta/ProductPageMeta.graphql
```diff
@@ -1,7 +1,26 @@
...
+  price_range {
+    maximum_price {
+      final_price {
+        value
+        currency
+      }
+      regular_price {
+        value
+      }
+    }
```
paales, line 17: Is there a reason we're using maximum_price here? In the case of a configurable with multiple prices this will return the higher price. Is that by design?

packages/magento-store/PageMeta.tsx
```diff
@@ -7,10 +7,11 @@ import { StoreConfigDocument } from './StoreConfig.gql'
 
 type PageMetaProps = Pick<NextPageMetaProps, 'title' | 'metaDescription' | 'metaRobots'> & {
   canonical?: string
+  children?: React.ReactNode
```
paales, line 10: I don't think this isn't necessary, as it is also included in NextpageMetaProps

examples/magento-graphcms/pages/[...url].tsx
```diff
@@ -75,7 +75,9 @@ function CategoryPage(props: CategoryProps) {
         metaDescription={page?.metaDescription}
         metaRobots={page?.metaRobots.toLowerCase().split('_') as MetaRobots[]}
         canonical={page?.url ? `/${page.url}` : undefined}
-        {...category}
+        name={category?.name}
+        meta_title={category?.meta_title}
+        meta_description={category?.meta_description}
```
Comment (paales) [inline examples/magento-graphcms/pages/[...url].tsx:80]: The CategoryMeta component should `Omit<PageMetaProps, 'title' | 'metaDescription' | 'children'>`. Then we can keep the `{...category}`

(It is a bit unfortunate that Magento opted to use `children` as the field which makes handling these situations a bit weird.)

paales:
It is starting to look really good @KMalkowski! Almost ready to merge IMO, right on time for GraphCommerce 6!

packages/magento-product/components/ProductPageMeta/ProductPageMeta.graphql
KMalkowski: I thought it would be safer to return the highest value here as min price could sometimes be a part of the product given bundles for example but in the end, it comes down to the kind of products a shop sells. Now I think the minimum price would be a more universal solution as many systems advertise the price in the "starting from" format. 
Done

examples/magento-graphcms/pages/[...url].tsx
Comment (KMalkowski) [reply examples/magento-graphcms/pages/[...url].tsx]: Yep, that's better, omitting children was enough 👌

packages/magento-store/PageMeta.tsx
KMalkowski: Done

packages/magento-category/components/CategoryMeta/CategoryMeta.tsx
```diff
@@ -5,17 +5,13 @@ import { i18n } from '@lingui/core'
 import { CategoryMetaFragment } from './CategoryMeta.gql'
 
 export type CategoryMetaProps = CategoryMetaFragment &
-  Omit<PageMetaProps, 'title' | 'metaDescription'> & {
+  Omit<PageMetaProps, | 'children'> & {
     params?: ProductListParams
-    title?: string | null
-    metaDescription?: string | null
```
KMalkowski, line 11: Reverted this change as these types are not compatible with NextPageMetaProps types where title and metaDescription can't be null

packages/magento-product/components/ProductPageMeta/ProductPageMeta.graphql
paales: I think that is a good solution :)

paales:
I've rebased on canary+force pushed. If everything builds, i'll merge

paales:
@KMalkowski Merged!! 🎉

paales:
I was thinking maybe as a followup:

- Might it be an idea to  move the og:url, og:title and og:image to the next-ui PageMeta?
- Disable the og:tags by default and enable with a configuration `enableOgtags: Boolean` or something.

ErwinOtten:
@KMalkowski This could be cool: https://vercel.com/docs/concepts/functions/edge-functions/og-image-generation/og-image-examples

KMalkowski:
> I was thinking maybe as a followup:
> 
> * Might it be an idea to  move the og:url, og:title and og:image to the next-ui PageMeta?
> * Disable the og:tags by default and enable with a configuration `enableOgtags: Boolean` or something.

Actually yeah, this could be useful for blog posts for example.

PR #1824, opened 2023-03-01
Fix issue where the discount value displayed in the cart gets the tax…

SimonPrins:
… subtracted twice when displaying prices excluding tax.

### DIFF
--- a/packages/magento-cart/components/CartTotals/CartTotals.tsx
+++ b/packages/magento-cart/components/CartTotals/CartTotals.tsx
@@ -86,12 +86,7 @@ export function CartTotals(props: CartTotalsProps) {
       )}
 
       {prices?.discounts?.map((discount) => {
-        const value = inclTax
-          ? (discount?.amount.value ?? 0) * -1
-          : (discount?.amount.value ?? 0) *
-            ((prices.subtotal_excluding_tax?.value ?? 1) /
-              (prices.subtotal_including_tax?.value ?? 1)) *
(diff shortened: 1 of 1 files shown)


paales:
Mmm https://github.com/graphcommerce-org/graphcommerce/commit/0d411350e4621928411c2800be6ea02c6125049a#diff-672ad1e5d7eb4b752dd8a266635a1b092a3b7a3de5f3644df8aab7ec0257c64fR110

SimonPrins:
This issue happens when you have both Magento and GC configured to display prices excluding taxes.
Magento will provide the discount excluding VAT, but then GC also tries to subtract the VAT and the result is an incorrect value.

I am not entirely certain whether the prices incl/excl tax setting for GC is expected to always match the Magento setting, or whether you are supposed to be able to use a different setting for GC than you have for Magento. I suppose it makes sense to be able to vary this, for instance if you have a single Magento install with multiple GC frontends for B2B and B2C.

If the Magento and GC settings are not required to match then this PR will not actually fix the issue, it will only move the problem to installs where Magento is configured for prices including tax and GC is configured to display prices excluding tax.

To fix the issue in both situations I think you would need to have additional information from Magento, for instance if you had both the discount including tax and the discount excluding tax then you can choose which you need. If you could find out which setting Magento is using that would also be helpful in determining whether or not GC needs to subtract VAT.

paales:
Not sure what to do with this?

PR #1826, opened 2023-03-01
Feature/spread val

ErwinOtten:


ErwinOtten:
After consultation, it has been concluded that these changes add to much churn and offer too little added value

PR #1843, opened 2023-03-14
Migrate away from direct @mui/material imports and instead use second level imports to improve performance.

paales:


paales:
Experiment doesn't result significant benefits.

PR #1858, opened 2023-03-21
Feature/gcom 1005

Giovanni-Schroevers:


examples/magento-graphcms/components/Layout/LayoutNavigation.tsx
```diff
@@ -128,7 +130,20 @@ export function LayoutNavigation(props: LayoutNavigationProps) {
...
+                        aria-label={i18n._(/* i18n */ 'Search...')}
+                        src={iconSearch}
+                        size='large'
+                        sx={{ color: 'text.primary' }}
+                      />
+                    </Fab>
+                  ) : (
+                    <SearchLink href='/search' aria-label={i18n._(/* i18n */ 'Search...')} />
+                  )}
+                </>
```
paales, line 146: - useMediaQuery will result in a CLS so we can not use this, we should solve this with CSS.
- Can we move this logic into SearchLink instead of the examples directory?

paales (changes requested):

.changeset/perfect-bananas-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphcommerce/demo-magento-graphcommerce': minor
```
paales, line 2:
```suggestion
'@graphcommerce/demo-magento-graphcommerce': patch
```

paales (changes requested):

examples/magento-graphcms/components/Layout/LayoutNavigation.tsx
Giovanni-Schroevers: Changing the component would also result in it changing in the sidebar, would adding a prop for this be desired?

Giovanni-Schroevers: And could you explain why using css instead of the mediaquery would not result in cls?

paales: > Changing the component would also result in it changing in the sidebar, which i assume is not desired. Would adding a prop for this be desired?

```
breakpoint?: Breakpoint
```

Like implemented in [LinkOrButton](https://github.com/graphcommerce-org/graphcommerce/blob/canary/packages/next-ui/Button/LinkOrButton.tsx)

PR #1862, opened 2023-03-22
Index

ErwinOtten:


docs/getting-started/create.md
```diff
@@ -1,148 +1,91 @@
...
-project, you need to update variables in the graphcommerce.config.js file. The
-graphcommerce.config.js file contains useful information about your storefront.
+## Step 3: Start the app
 
-Create a copy: `cp graphcommerce.config.js.example graphcommerce.config.js` and
-configure the following fields:
+```bash
+yarn
+# Install the dependencies
+```
```
paales, line 68: Lets do this in Step 1.

docs/getting-started/create.md
```diff
@@ -1,148 +1,91 @@
...
+cp -R graphcommerce/examples/magento-graphcms/. my-project && rm -rf graphcommerce && cd my-project
+# Copy example, delete repo, navigate to project folder
+```
 
-  <figcaption>GraphCommerce installation</figcaption>
-</figure>
+## Step 2: Configure API keys
 
-### Download the example
+Rename the configuration example file to: `graphcommerce.config.js` and
```
paales, line 35: If you do not have a graphcommerce.config.js file in your project it will start in demo mode and everything should be running as expected.

paales: Suggest copying the example instead of renaming? `cp graphcommerce.config.js.example graphcommerce.config.js`

docs/readme.md
```diff
@@ -30,54 +30,135 @@ https://user-images.githubusercontent.com/1251986/154977573-4015e77c-43e7-481e-a
...
+### Build your custom storefront
 
-- Typescript
-- React
-- Nextjs
-- GraphQL
-- Mui
-- Apollo Client
-- GraphQL Code Generator
-- Framer Motion
```
paales, line 42: We're removing this here, should we find another place, this is one of the things mentioned by customers that they were looking for this stack.

docs/getting-started/create.md
ErwinOtten: fixed

docs/readme.md
ErwinOtten: Yes this will be moved to another location

PR #1870, opened 2023-03-24
feat(GCOM-1029): preload crosssells when adding items to the cart to …

bramvanderholst:
…prevent jank in 'added' overlay

Dependant on https://github.com/graphcommerce-org/graphcommerce/pull/1869

### DIFF
--- a/packages/magento-product/components/AddProductsToCart/AddProductsToCartForm.tsx
+++ b/packages/magento-product/components/AddProductsToCart/AddProductsToCartForm.tsx
@@ -1,5 +1,6 @@
 import { UseFormGraphQlOptions } from '@graphcommerce/ecommerce-ui'
-import { useFormGqlMutationCart } from '@graphcommerce/magento-cart'
+import { useApolloClient } from '@graphcommerce/graphql'
+import { useFormGqlMutationCart, CrosssellsDocument } from '@graphcommerce/magento-cart'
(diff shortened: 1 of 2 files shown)


packages/magento-product/components/AddProductsToCart/AddProductsToCartForm.tsx
```diff
@@ -80,6 +82,19 @@ export function AddProductsToCartForm(props: AddProductsToCartFormProps) {
...
+
+      if (sku) {
+        // Preload crosssells
+        await client.query({
+          query: CrosssellsDocument,
+          variables: { pageSize: 1, filters: { sku: { eq: sku } } },
+        })
+      }
+
+      return requestData
```
paales, line 97: We are now waiting for the crossells query to be completed _before_ we are submitting the form. What we want:
1. Fire of the Crosssells query in onBeforeSubmit
2. Wait for the result in of the query in onComplete.

You can probably store the promise in a useRef.

paales: And second, we only want to do this `if(redirect="added")`

bramvanderholst: @paales ✅ Fixed

PR #1871, opened 2023-03-24
Feature/checkout checkboxes

bramvanderholst:


packages/ecommerce-ui/components/FormComponents/CheckboxElement.tsx
```diff
@@ -55,7 +61,7 @@ export function CheckboxElement<TFieldValues extends FieldValues>({
                     {...rest}
                     color={rest.color || 'primary'}
                     sx={{
-                      ...rest.sx,
+                      ...sx,
```
paales, line 64: I don't think this'll work (not introduced in your change, just seeing it now):  https://mui.com/system/getting-started/the-sx-prop/#passing-the-sx-prop

```
        // You cannot spread `sx` directly because `SxProps` (typeof sx) can be an array.
        ...(Array.isArray(sx) ? sx : [sx]),
```

packages/ecommerce-ui/components/FormComponents/CheckboxElement.tsx
```diff
@@ -12,12 +12,16 @@ import {
...
+  SxProps,
+  Theme,
 } from '@mui/material'
 
 export type CheckboxElementProps<T extends FieldValues> = Omit<CheckboxProps, 'name'> & {
   parseError?: (error: FieldError) => string
   label?: FormControlLabelProps['label']
   helperText?: string
+  sx?: SxProps<Theme>
+  sxFormControl?: SxProps<Theme>
```
paales, line 24: Might it be an idea to do something like:

```
formControl?: Omit<FormControlProps<'div'>, 'required' | 'error'>
```

paales (changes requested):

packages/ecommerce-ui/components/FormComponents/CheckboxElement.tsx
bramvanderholst: Good idea! Done :)

PR #1873, opened 2023-03-29
fix(GCOM-1028): allow cache header to be set in config, change next-data cache handler

Giovanni-Schroevers:


examples/magento-graphcms/next.config.js
```diff
@@ -3,9 +3,26 @@
...
+      options: {
+        cacheName: 'next-data',
+        expiration: {
+          maxEntries: 32,
+          maxAgeSeconds: 24 * 60 * 60, // 24 hours
+        },
+      },
+    },
+    ...cache,
+  ],
```
paales, line 25: All changes made in the examples/magento-graphcms directory create 'churn' for the integrator when we need to make changes to the caching strategy in the future.

Might it be an idea to export a better runtimeCaching from `@graphcommerce/next-config`?

examples/magento-graphcms/next.config.js
```diff
@@ -25,6 +42,58 @@ const nextConfig = {
...
+      headers: [
+        {
+          key: 'Cache-Control',
+          value: 'private, no-cache, no-store, max-age=0, must-revalidate',
+        },
+      ],
+
+      source: '/_next/:path(.+\\.(?:json)$)',
+    },
+  ],
```
paales, line 96: What does this do, this seems a change of a _very_ large scope by modifying all headers for all requests?

paales (changes requested):

PR #1880, opened 2023-03-29
Fix/gcom 1028

Giovanni-Schroevers:
# What files should / shouldn't be cached

### next/static
"Next.js automatically adds caching headers to immutable assets served from /_next/static including JavaScript, CSS, static images, and other media." from https://nextjs.org/docs/going-to-production#caching

### next/data
These can include dynamic data like products, and should thus not be cached. 

### HTML Document
Has references to hashed static files when these hashes are updated the document would be out of date, so should not be cached.


paales (changes requested):

paales:
It seems all category routes do not get thrown in the `next-data` cache but end up in the `others` cache. The `others` cache already has NetworkFirst, so this should work as expected. 

Scherm­afbeelding 2023-04-06 om 13 31 39

Scherm­afbeelding 2023-04-06 om 13 32 02

So the change of the service worker is not necessary because it already works as expected.

The modifications to the headers is too experimental in my opinion, so that doesn't leave anything to do in this PR. What do you think @Giovanni-Schroevers?

Giovanni-Schroevers:
@paales There are still some next-data urls that get cached but you could make the argument that these are not that important and its fine to be served from the cache for a day. 
Screenshot 2023-04-07 at 11 12 16
So I will go ahead and close this.

PR #1885, opened 2023-03-31
chore: use correct path for additionalResolvers in meshrc.yml

carlocarels90:


packages/address-fields-nl/lib/nationaalgeoregister/.meshrc.yaml
```diff
@@ -10,4 +10,4 @@ sources:
             method: GET
             responseSample: 'lib/nationaalgeoregister/responseSample.json'
 additionalResolvers:
-  - 'lib/nationaalgeoregister/nationaalgeoregister.ts'
+  - lib/nationaalgeoregister/resolver
```
paales, line 13: Can we check if this works as well?

```suggestion
  - @graphcommerce/address-fields-nl/lib/nationaalgeoregister/resolver
```

If that works, we can update the README as well that we do not need to copy the file

PR #1889, opened 2023-03-31
Put the country field first, so the address fields will not be changed afterwards when postcode service is active.

carlocarels90:


paales:
Do we only want to do this when the addressfields plugin is active? Can we perhaps add a prop to the AddressFields `countryFirst?: boolean` and set that to true in the plugin?

carlocarels90:
@paales I've created the optional `countryFirst` boolean. When this is true, the country field will be on top of de AddressField fields.

packages/address-fields-nl/components/PostcodeNLAddressFields.tsx
```diff
@@ -47,6 +47,44 @@ export function PostcodeNLAddressFields(props: AddressFieldsProps) {
...
+                readOnly,
+                endAdornment: <InputCheckmark show={valid.regionId} select />,
+              }}
+              options={filterNonNullableKeys(regionList, ['id', 'name']).map(
+                ({ id, name: label }) => ({ id, label }),
+              )}
+            />
+          )}
+        </FormRow>
+      )}
```
paales, line 87: This whole block is the same as below, so you can just assign it to a variable:


```tsx
const countryFields = (
  <FormRow>
    ...
  </FormRow>
)
```

And conditionally render it;

```suggestion
      {countryFirst && countryFields}
```

PR #1900, opened 2023-04-04
chore: make the SelectElement validation message translatable

carlocarels90:


paales:
@carlocarels90 Can you update all other *Element components as well?

PR #1932, opened 2023-05-12
Data fetching app dir

paales:


paales:
I've communicated on the slack chat earlier:

> The current appDir implementation is highly unstable and requires:
> - Patching @types/react
> - Running alpha branches of typescript
> - Running canary branches of nextjs
> 
> It is very slow to build at the moment (about 4x as slow). Experiencing constant OOM errors, running into obscure internal react errors etc.
> 
> We’re relying heavily on Apollo Client, they are one of the first libraries that take RSC seriously ([apollo-client-nextjs](https://github.com/apollographql/apollo-client-nextjs)), but it still is very incomplete at the moment.
> 
> I think that marking the app dir as stable was more of a point to communicate to developers: We need feedback, we’re not getting it while it is in beta, since nobody is using it. Library maintainers: It’s your turn, this API is what you are going to get.
>
> We do however got streaming rendering working, which is exciting!
>
> As for the future migration story: We’ll either keep supporting the pages directory for the years to come, or we’ll provide a smooth migration path in the future.

PR #1936, opened 2023-05-16
Added redirect for /customer/account so links in (transactional) emails function properly

bramvanderholst:


examples/magento-graphcms/next.config.js
```diff
@@ -25,6 +25,10 @@ const nextConfig = {
   eslint: {
     ignoreDuringBuilds: true,
   },
+  // eslint-disable-next-line @typescript-eslint/require-await
+  async redirects() {
+    return [{ source: '/customer/account', destination: '/account', permanent: true }]
+  },
```
paales, line 31: Can we move this redirect to this location? https://github.com/graphcommerce-org/graphcommerce/blob/canary/packagesDev/next-config/src/withGraphCommerce.ts#L94

Since we're modifying a next-config package we also need to compile those by running `yarn packages` in the root followed by a `yarn graphcommerce codegen-config` in the examples/magento-graphcms directory.

bramvanderholst: @paales Like this? https://github.com/graphcommerce-org/graphcommerce/pull/1936/commits/57ea9a489ea69d6db7e0aee17fb4bfe8a52272d7

PR #1938, opened 2023-05-17
Feature/gcom 1089  - Hygraph migrations

JoshuaS98:
### TEST
This can be tested as described below in the "Migrations test" project in Hygraph under Reach Digitals account.

To test, please use migration 6.2. If there are no changes, try to remove dynamic row model from the schema and run the migration again. 

### DOCUMENTATION
**1. Install the @graphcommerce/hygraph-cli package**

**2. Create a Hygraph migration token**

Open your hygraph project > Project settings > Permanent auth tokens

Click  'Add token' and give it a name, something like 'GraphCommerce Write Access Token' and keep stage on 'Published'.

Under 'Management API', click 'Yes, Initialize defaults'

Click 'Edit Permissions' and enable: 'Update' and 'Delete' permissions for 'models', 'enumerations', 'fields', 'components' and 'sources'
    - Update existing models
    - Delete existing models
    - Update existing fields
    - Delete existing fields
    - Update existing enumerations
    - Delete existing enumerations
    - Update existing components
    - Delete existing components
    - Update remote sources
    - Delete remote sources
    - Read existing environments
    - Read public content views
    - Create public content views
    - Update public content views
    - Delete public content views
    - Can see schema view

**Add this new token to your env file:** 
GC_HYGRAPH_WRITE_ACCESS_ENDPOINT=””

**3. Add the Content API key to your env file:** 
GC_HYGRAPH_WRITE_ACCESS_ENDPOINT="https://api-eu-central-1-shared-euc1-02.hygraph.com/v2/.../master"

**4. Add your projectID to your env file**
GC_HYGRAPH_PROJECT_ID=””

**5. Run yarn graphcommerce hygraph-migrate**

**6. Select the migration you want to run and press enter.** 

**7. The migrations should now be set, check in Hygraph Schema.**


packages/hygraph-cli/dist/functions.js
```diff
@@ -0,0 +1,181 @@
...
+        warning: '\x1b[33m\x1b[1m%s\x1b[0m',
+        info: '\x1b[36m\x1b[1m%s\x1b[0m',
+    };
+    // eslint-disable-next-line no-console
+    console.log(type ? color[type] : '', `[GraphCommerce]: ${message}`);
+};
+exports.graphcommerceLog = graphcommerceLog;
+const migrationAction = (client, schema, type, action, props, parentApiId, parentType) => {
+    const alreadyExists = () => {
+        if (type === 'model') {
```
mikekeehnen, line 18: You could use a switch for type instead of running multiple if statements

packages/hygraph-cli/src/functions.ts
```diff
@@ -0,0 +1,315 @@
...
+  const validProp = validPropMap[type] && validPropMap[type][action]
+  const actionFunc = actionMap[action] && actionMap[action][type]
+
+  if (!alreadyExists()) {
+    if (validProp && actionFunc) {
+      graphcommerceLog(`${capitalize(action)} ${type} with apiId ${props.apiId}...`)
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore | This error is a loss on typescript autocomplete, but the function is called correctly
+      actionFunc(props)
+    }
```
mikekeehnen, line 311: Might want to log which prop isn't valid for future debugging

docs/upgrading/graphcommerce-6-1-to-6-2.md
```diff
@@ -87,3 +87,67 @@ In your package.json
...
+
+      - Can see schema view
+
+   5. Add this new token to your env file like this:
+      `GC_HYGRAPH_WRITE_ACCESS_ENDPOINT=””`
+
+3. Add the Content API key to your env file like this:
+   `GC_HYGRAPH_WRITE_ACCESS_ENDPOINT="https://api-eu-central-1-shared-euc1-02.hygraph.com/v2/.../master"`
+4. Add your projectID to your env file like this: `GC_HYGRAPH_PROJECT_ID=””`
+5. Run `yarn graphcommerce hygraph-migrate`
```
mikekeehnen, line 150: It might be usefull to clearly state that the user can use both the `graphcommerce.config.js` file and the `.env` file to configure these values. When I looked through your PR, I got a little confused because both were used.

mikekeehnen (approved):

docs/upgrading/graphcommerce-6-1-to-6-2.md
```diff
@@ -87,3 +87,67 @@ In your package.json
...
+## Upgrading your Hygraph schema
+
+In version 6.2 and onwards it is possible to upgrade your Hygraph schema using
+the Graphcommerce Hygraph CLI. This is a tool that automatically modifes your
+Hygraph schema without the need of adding models/components/enumarations
+manually. The following steps are needed to utilize this tool:
+
+1. Install the @graphcommerce/hygraph-cli package
+
+2. Create a Hygraph migration token
```
paales, line 100: I think it might be best to create a separate Hygraph documentation page, something like `docs/hygraph/README.md` where we write about hygraph where the only documentation right now is how to use the CLI.

These upgrade instructions should specify which migrations are expected to be ran:

-  Use the [Hygraph CLI] to apply migrations x and y.

docs/upgrading/graphcommerce-6-1-to-6-2.md
```diff
@@ -87,3 +87,67 @@ In your package.json
...
+      - Can see schema view
+
+   5. Add this new token to your env file like this:
+      `GC_HYGRAPH_WRITE_ACCESS_ENDPOINT=””`
+
+3. Add the Content API key to your env file like this:
+   `GC_HYGRAPH_WRITE_ACCESS_ENDPOINT="https://api-eu-central-1-shared-euc1-02.hygraph.com/v2/.../master"`
+4. Add your projectID to your env file like this: `GC_HYGRAPH_PROJECT_ID=””`
+5. Run `yarn graphcommerce hygraph-migrate`
+6. Select the migration you want to run and press enter.
```
paales, line 151: Which upgrade does a user want to run?

Should we also communicate to the user that an upgrade can't be rolled back so they should probably test with a copy of the current Hygraph installation?

graphql.config.js
```diff
@@ -1,7 +1,8 @@
 /**
  * GraphQL Configuration
  *
- * [GraphQL VSCode exstension](https://marketplace.visualstudio.com/items?itemName=GraphQL.vscode-graphql)
+ * [GraphQL VSCode
+ * exstension](https://marketplace.visualstudio.com/items?itemName=GraphQL.vscode-graphql)
```
paales, line 5: Newline shouldn't be here.

docs/upgrading/graphcommerce-6-1-to-6-2.md
```diff
@@ -87,3 +87,67 @@ In your package.json
...
+
+      - Read public content views
+
+      - Create public content views
+
+      - Update public content views
+
+      - Delete public content views
+
+      - Can see schema view
```
paales, line 142: This seems a very long list. Is there also an 'admin' token that has access to everything or is the user required to select from the list? Might be a suggestion that a user either can create a token with these actions or an admin token.

Also mention that the token is _only_ required during a migration and nowhere else.

docs/upgrading/graphcommerce-6-1-to-6-2.md
```diff
@@ -87,3 +87,67 @@ In your package.json
...
+
+   5. Add this new token to your env file like this:
+      `GC_HYGRAPH_WRITE_ACCESS_ENDPOINT=””`
+
+3. Add the Content API key to your env file like this:
+   `GC_HYGRAPH_WRITE_ACCESS_ENDPOINT="https://api-eu-central-1-shared-euc1-02.hygraph.com/v2/.../master"`
+4. Add your projectID to your env file like this: `GC_HYGRAPH_PROJECT_ID=””`
+5. Run `yarn graphcommerce hygraph-migrate`
+6. Select the migration you want to run and press enter.
+7. The migrations should now be upgraded, check your Hygraph Schema if changes
```
paales, line 152: We should validate and make absolutely clear that Hygraph migrations are backwards compatible so they can be applied on an existing installation.

paales (changes requested):

PR #1942, opened 2023-05-26
Fix/session issues

paales:


packages/graphql/components/GraphQLProvider/createCacheReviver.ts
```diff
@@ -23,24 +24,26 @@ export async function createCacheReviver(
     try {
       const { cache } = client
 
-      if (persistor) await persistor.persist()
-      // todo https://github.com/apollographql/apollo-cache-persist/tree/master/examples/react-native/src/utils/persistence
+      if (persistor) {
+        await persistor.persist()
+        return
+      }
```
paales, line 30: During the development of the persistenceMapper we detected that the cacheReviver would recreate the cache lots of times. It now doesn't do that as much anymore by bailing early here.

packages/magento-customer/hooks/useCustomerValidateToken.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  useCustomerQuery(UseCustomerValidateTokenDocument, {
+    initialFetchPolicy: 'network-only',
+    onError: async ({ graphQLErrors }) => {
+      const category: ErrorCategory = 'graphql-authorization'
+      // If there is no authorization error, do nothing.
+      if (!graphQLErrors.some((e) => e.extensions?.category === category)) return
+
+      await client.clearStore()
+      router.reload()
```
paales, line 20: I'm not really happy this is required, but cleanStore will refetch all queries which isn't what we want because most of those queries will then fail.

Ideas welcome

PR #1952, opened 2023-06-01
chore: enable password fields when there is an error and user input correction is required

carlocarels90:
Before:
image

After:
image

### DIFF
--- a/.changeset/fresh-icons-wait.md
+++ b/.changeset/fresh-icons-wait.md
@@ -1,0 +1,5 @@
+---
+'@graphcommerce/magento-customer': patch
+---
+
+enable password fields when there is an error and user input correction is required.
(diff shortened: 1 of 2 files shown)


paales:
@carlocarels90 I can not reproduce the issue on the demo and from a code perspective I don't think this should be necessary. Can you help me reproduce it?

carlocarels90:
@paales 

### Steps to reproduce:

- Go to https://graphcommerce.vercel.app/ (Not logged in).
- Add a product to your cart (e.g., https://graphcommerce.vercel.app/p/subtle-sympathy-gc-1087-sock).
- Proceed to the Checkout Shipping page and use an email address that does not yet exist in GraphCommerce (https://graphcommerce.vercel.app/checkout).
- Fill in all the other required fields and place the order.
- On the Success page, scroll to "_No account yet?_" and click on "_Create an account_".
- In the password fields, enter `22222222` and click on "_Create Account_".
- Now you are not allowed to change/update the password fields.

PR #1959, opened 2023-06-09
Feature/gcom 1101 | Sidebar filters

JoshuaS98:


examples/magento-graphcms/pages/[...url].tsx
```diff
@@ -68,6 +54,8 @@ function CategoryPage(props: CategoryProps) {
   const isLanding = category?.display_mode === 'PAGE'
   const page = pages?.[0]
   const isCategory = params && category && products?.items && filterTypes
+  const theme = useTheme()
+  const matches = useMediaQuery(theme.breakpoints.up('md'))
```
Comment (paales) [inline examples/magento-graphcms/pages/[...url].tsx:58]: `useMediaQuery` is not allowed to be used to determine if elements should be shown on the frontend. Use CSS to determine if things should be shown.

useMediaQuery uses Javascript to determine this and thus will create a Layout shift when loading either on mobile or desktop.

examples/magento-graphcms/pages/[...url].tsx
```diff
@@ -111,46 +99,17 @@ function CategoryPage(props: CategoryProps) {
...
+            mode={matches ? 'sidebar' : 'default'}
+            ProductListItems={ProductListItems}
+            ProductListCount={ProductListCount}
+            ProductListPagination={ProductListPagination}
+            products={products}
+            filters={filters}
+            params={params}
+            filterTypes={filterTypes}
+            category={category}
+          />
```
Comment (paales) [inline examples/magento-graphcms/pages/[...url].tsx:112]: Can we first try to build this without creating this abstraction?

Might be interesting to create a FilterLayout which is _only_ a presentation layout? Also the Layout would only apply for the ProductFiltersPro of course.

packages/magento-product/components/ProductFiltersPro/ProductFiltersPro.tsx
```diff
@@ -32,14 +37,14 @@ export const useProductFiltersPro = () => {
 export type FilterFormProviderProps = Omit<
   UseFormProps<ProductFilterParams>,
   'values' | 'defaultValues'
-> & { children: React.ReactNode; params: ProductListParams }
+> & { children: React.ReactElement; params: ProductListParams }
```
paales, line 40: Is this really necessary?

packages/next-ui/ActionCard/ActionCard.tsx
```diff
@@ -122,7 +128,7 @@ export function ActionCard(props: ActionCardProps) {
             theme.shape.borderRadius * 3,
             theme.breakpoints.values,
           ),
-
+          display: index + 1 > maxItems && !show ? 'none!important' : 'block',
```
paales, line 131: Can we find a better solution for this, I don't like the `!important` here.

packages/next-ui/ActionCard/ActionCard.tsx
```diff
@@ -96,6 +99,9 @@ export function ActionCard(props: ActionCardProps) {
     variant = 'outlined',
     layout = 'list',
     error = false,
+    index = 0,
+    show = true,
+    maxItems = 100,
```
paales, line 104: These props don't belong here, this should be configured on the ActionCardList.

packages/next-ui/ActionCard/ActionCardListForm.tsx
```diff
@@ -51,29 +56,53 @@ export function ActionCardListForm<
...
+              <IconSvg
+                sx={{
+                  transform: show ? 'rotate(180deg)' : 'rotate(0deg)',
+                  transition: 'transform 0.3s ease-in-out',
+                }}
+                src={iconChevronDown}
+              />
+            </Button>
+          )}
+        </>
```
paales, line 105: I'm going to be annoying here, what if the currently selected item is in the list you've collapsed, shouldn't it still be shown?

packages/next-ui/ActionCard/ActionCardListForm.tsx
```diff
@@ -36,6 +39,8 @@ export function ActionCardListForm<
     ...other
   } = props
   const RenderItem = render as React.FC<ActionCardItemRenderProps<ActionCardItemBase>>
+  const [show, setShow] = React.useState(false)
+  const maxItems = 4
```
paales, line 43: Might it be an idea to move this to the underlying ActionCardList instead of ActionCardListForm? Also, it should have `showMoreAfter={4}` prop or something which isn't configured by default.

Maybe even ActionCardLayout?

packages/magento-product/components/ProductFiltersPro/FilterLayout.tsx
```diff
@@ -0,0 +1,131 @@
...
+    filters,
+    params,
+    category,
+    mode = 'default',
+    sx = [],
+  } = props
+
+  if (mode === 'sidebar' && import.meta.graphCommerce.productFiltersPro) {
+    return (
+      <Container maxWidth='lg' sx={Array.isArray(sx) ? sx : [sx]}>
```
paales, line 51: We do not want to limit the width of the page, if a user can configure this, that might be interesting.

packages/magento-product/components/ProductFiltersPro/FilterLayout.tsx
```diff
@@ -0,0 +1,131 @@
...
+        <Box
+          sx={(theme) => ({
+            display: 'grid',
+            gridTemplateColumns: '3fr 9fr',
+            columnGap: theme.spacings.md,
+            [theme.breakpoints.down('md')]: { display: 'flex' },
```
paales, line 57: - Use responsive object values whenever possible: `display: { xs: 'flex', md: 'flex'}`
- Please develop mobile first from a code perspective, don't use `down`, use `up('md')` whenever possible.
- Do not switch rendering technique for mobile / desktop. Choose grid or flex for both

packages/magento-product/components/ProductFiltersPro/ProductFilterAccordion.tsx
```diff
@@ -0,0 +1,69 @@
...
+                  marginBottom: 0,
+                  '& .SectionHeader-left': {
+                    color: theme.palette.text.primary,
+                    fontSize: theme.typography.body1.fontSize,
+                  },
+                  '& .SectionHeader-right > button': {
+                    padding: 0,
+                    margin: 0,
+                    fontSize: theme.typography.body1.fontSize,
+                  },
```
paales, line 52: We must be able to do better here, this code looks gross. Please reconsider the strategy used.

packages/magento-product/components/ProductFiltersPro/ProductFilterRangeSection.tsx
```diff
@@ -25,38 +26,51 @@ export function ProductFilterRangeSection(props: FilterProps) {
...
-                    <Trans id='Clear' />
-                  </Button>
-                ) : undefined
+    <Box
+      sx={(theme) => ({
+        [theme.breakpoints.down('md')]: {
+          display: 'none',
+        },
+      })}
+    >
```
paales, line 35: - We're adding two additional divs in this component, that seems wasteful.
- Visbility should not be controlled here, but by one the parent components.

packages/magento-product/components/ProductFiltersPro/ProductFilterRangeSection.tsx
```diff
@@ -25,38 +26,51 @@ export function ProductFilterRangeSection(props: FilterProps) {
...
+        render={({ field: { onChange, value } }) => {
+          if (value && !isFilterTypeRange(value)) throw new Error('Invalid filter type')
+
+          const from = value?.from ? Number(value?.from) : min
+          const to = value?.to ? Number(value?.to) : max
+
+          return (
+            <ProductFilterAccordion
+              summary={
+                <Box sx={{ my: 2 }}>
```
paales, line 48: The my: 2 can probably be moved to a prop of the accordion it's self.

packages/next-ui/ActionCard/ActionCardListForm.tsx
paales: I do like the simple display block / none strategy here, keeps it simple for now.

packages/magento-product/components/ProductFiltersPro/ProductFilterRangeSection.tsx
```diff
@@ -25,38 +26,51 @@ export function ProductFilterRangeSection(props: FilterProps) {
...
+                  <SectionHeader
```
paales, line 49: Maybe we can drop the SectionHeader completely as we're using the accordion now? An accordion also allows additional items on the right.

paales (changes requested):

examples/magento-graphcms/pages/[...url].tsx
```diff
@@ -89,9 +87,9 @@ function CategoryPage(props: CategoryProps) {
           <LayoutTitle
             variant='h1'
             gutterTop
-            sx={(theme) => ({
+            sx={{
               marginBottom: category?.description && theme.spacings.md,
-            })}
+            }}
```
Comment (paales) [inline examples/magento-graphcms/pages/[...url].tsx:92]: This can be reverted.

examples/magento-graphcms/pages/search/index.tsx
```diff
@@ -14,20 +11,22 @@ import {
...
   ProductFiltersQuery,
-  ProductFiltersProAllFiltersChip,
-  ProductFiltersProLimitChip,
   AddProductsToCartForm,
+  FilterLayout,
+  ProductListCount,
+  ProductListFilters,
+  ProductListItems,
+  ProductListPagination,
+  ProductListSort,
```
paales, line 20: I don't this is correct. The ProductListCount**Search** is by design, so it can be overwritten specifically for the search page.

examples/magento-graphcms/pages/search/index.tsx
```diff
@@ -38,6 +37,7 @@ import {
   GetStaticProps,
   LayoutTitle,
   LayoutHeader,
+  responsiveVal,
```
paales, line 40: Not allowed in the examples directory

examples/magento-graphcms/pages/search/index.tsx
```diff
@@ -56,6 +56,21 @@ export type GetPageStaticProps = GetStaticProps<
...
+    <AddProductsToCartForm>
+      <ProductListItemsSearch
+        renderers={productListRenderer}
+        title={title}
+        items={items}
+        loadingEager={1}
+      />
+    </AddProductsToCartForm>
+  )
+}
```
paales, line 72: Not allowed in the examples directory

packages/next-ui/ActionCard/ActionCardListForm.tsx
```diff
@@ -36,6 +39,8 @@ export function ActionCardListForm<
     ...other
   } = props
   const RenderItem = render as React.FC<ActionCardItemRenderProps<ActionCardItemBase>>
+  const [show, setShow] = React.useState(false)
+  const showMoreAfter = 4
```
paales, line 43: showMoreAfter can not be hardcoded here and must be an opt-in functionality. This should be passed from the caller side `<ActionCardListForm showMoreAfter={4}/>`

paales (changes requested):
- The mobile sidebar filter now applies on click, instead when pressing apply. We should have more than one mode to handle this.

- I'd like to refactor the `<FilterLayout />` component to see if we can extract the layout concerns from the render concerns. Lets do this together.

paales (approved):

PR #1963, opened 2023-06-13
Added config to be able to hide the back button

bramvanderholst:


packages/next-ui/Config.graphqls
```diff
@@ -0,0 +1,6 @@
+extend input GraphCommerceConfig {
+  """
+  Hide the back button
+  """
+  hideBackButton: Boolean
```
paales, line 5: This sounds like it could just be a prop on `<LayoutHeaderBack/>` and passed through from `<LayoutHeader/>`?

bramvanderholst: LayoutHeader is also a next-ui component, so that would mean you'd have to pass it to LayoutHeader on every single page individually.

paales: Might it be an idea to:
- Create a prop to be able to hide the back button on LayoutHeader or something
- Create a local plugin

PR #1965, opened 2023-06-14
Added prop to LayoutHeader to be able to hide the back button

bramvanderholst:


packages/next-ui/Layout/components/LayoutHeader.tsx
```diff
@@ -54,7 +57,7 @@ export function LayoutHeader(props: LayoutHeaderProps) {
     sx = [],
     bgColor,
   } = props
-  const showBack = useShowBack()
+  const showBack = useShowBack() && !hideBackButton
```
paales, line 60: Do we want the back button to be shown on mobile?

PR #1970, opened 2023-06-22
Document restriction on locale format

hnsr:


hnsr:
Might actually be better if we could support `_` separators to allow keeping everything consistent with HyGraph locales (which doesn't allow `-` characters), and Magento store codes

I believe this is only a limitation of lingui currently, which can possibly be fixed in `examples/magento-graphcms/lingui.config.js`

paales:
@hnsr Next.js only accepts locales and doesn't support any non standard specification locale, right?

hnsr:
> @hnsr Next.js only accepts locales and doesn't support any non standard specification locale, right?

Yes, but it seems something goes wrong with finding translations when we use `_` as separators (which the spec should allow), which we tried to keep our HyGraph locales and storecodes the same.

However it looks like this may be something we do ourselves in https://github.com/graphcommerce-org/graphcommerce/blob/canary/packages/lingui-next/config.js#L19

I will close this PR and may open one later with an actual fix if I can make it work 👍🏻

paales:
@hnsr There are more locations where we extract the country or locale from the string, we should probably parse it better in those locations as well.

PR #1978, opened 2023-06-29
chore(GCOM-1074): Implement ConfigurableProductUrls component for dynamic URL handling

carlocarels90:


packages/magento-product-configurable/plugins/ConfigurableProductPageMeta.tsx
```diff
@@ -0,0 +1,48 @@
...
+    name,
+    meta_title,
+    meta_description,
+    url_key,
+    __typename,
+    ...rest
+  } = props
+  const { configured } = useConfigurableOptionsSelection({ url_key, index })
+
+  const configurableOption = configured?.configurable_product_options_selection?.variant
```
paales, line 31: Might something like this be an idea? `mergeDeep(props, variant)`?

paales (changes requested):

packages/magento-product-configurable/components/ConfigurableProductReviewChip/ConfigurableProductReviewChip.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const simpleProductRatingSummary =
+    configured?.configurable_product_options_selection?.variant?.rating_summary
+
+  return (
+    <ProductReviewChip
+      rating={simpleProductRatingSummary ?? rating_summary}
+      reviewSectionId='reviews'
+    />
+  )
+}
```
paales, line 23: Interesting component, not sure what to do with this? Should we always try and get the reviews from the configurable or does it make sense to get the reviews from the variant as well?

packages/magento-product-configurable/components/ConfigurableProductUrls/ConfigurableProductUrls.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const { configured } = useConfigurableOptionsSelection({ url_key, index })
+
+  useEffect(() => {
+    const currentUrlPath = router.asPath.split('/')
+    const currentUrlKey = currentUrlPath[currentUrlPath.length - 1]
+    const simpleProductUrlKey = configured?.configurable_product_options_selection?.variant?.url_key
+
+    if (simpleProductUrlKey && simpleProductUrlKey !== currentUrlKey) {
+      router
+        .replace(`/p/${simpleProductUrlKey}`, undefined, { scroll: false, shallow: true })
```
paales, line 22: There is a `productLink` method to get the correct route. I believe some projects are using a different route than `/p/[url]`.

packages/magento-product-configurable/components/ConfigurableProductUrls/ConfigurableProductUrls.tsx
```diff
@@ -0,0 +1,30 @@
...
+        .catch((error) => {
+          console.error('Failed to replace URL:', error)
```
paales, line 24: Any information when this is happening?

packages/magento-product-configurable/components/ConfigurableProductUrls/ConfigurableProductUrls.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export function ConfigurableProductUrls(props: ConfigurableProductUrlsProps) {
+  const { url_key, index = 0 } = props
+  const router = useRouter()
+  const { configured } = useConfigurableOptionsSelection({ url_key, index })
+
+  useEffect(() => {
+    const currentUrlPath = router.asPath.split('/')
+    const currentUrlKey = currentUrlPath[currentUrlPath.length - 1]
+    const simpleProductUrlKey = configured?.configurable_product_options_selection?.variant?.url_key
```
paales, line 18: How do we determine if the simple is visible individually? It can be that some products aren't visible in the Catalog, right?

packages/magento-product-configurable/plugins/ConfigurableProductPageMeta.tsx
paales: In general, I'm not even sure this is necessary. This information only matters when a page loads, not dynamically, right?

paales (changes requested):

packages/magento-product-configurable/plugins/ConfigurableProductPageMeta.tsx
carlocarels90: Yes, it's only necessary when the page loads, but if your simple products contain other metadata, you need to retrieve this data from the `configured?.configurable_product_options_selection?.variant`.

packages/magento-product-configurable/components/ConfigurableProductReviewChip/ConfigurableProductReviewChip.tsx
carlocarels90: Currently, all the reviews are always placed with the configurable product. This means that there are more reviews available to show. However, if the variants are very different from each other, it might be better to allow placing a separate review for each variant. I haven't decided yet on the best way to make this happen, maybe it could be an optional feature. What do you think?

packages/magento-product-configurable/components/ConfigurableProductUrls/ConfigurableProductUrls.tsx
carlocarels90: No, I see that we don't handle this anywhere else in the project either. So I think this one can be removed.

action-simon:
I have a question regarding PR 1978. Does it mean that the product url will be different for each variant and dynamically update when you click on it? On my previous Magento 2 project I have used something like that (from Best4Mage) and got into huge SEO problems because Google did not know what to do with these URLs resulting in duplicate content errors and so on. In the end I reverted it so that the url stays the same.

carlocarels90:
@action-simon 
> I have a question regarding PR 1978. Does it mean that the product url will be different for each variant and dynamically update when you click on it? On my previous Magento 2 project I have used something like that (from Best4Mage) and got into huge SEO problems because Google did not know what to do with these URLs resulting in duplicate content errors and so on. In the end I reverted it so that the url stays the same.

Hi Simon,

Yes, you understood it correctly. When you click on a different variant, the URLs will indeed change. However, we want to prevent duplicate content issues by implementing Canonical URLs. These Canonical URLs are applied in the `<ProductPageMeta />` component, which uses the URL of the Configurable Product. Search engines will index this version of the product, and all variants will be directed to it.

Additionally, it's important to note that this functionality will be optional. When this functionality is available you will be able to enable or disable this option according to your needs in your config file (`graphcommerce.config.js`).

Thank you for bringing it up!

Let me know if you have any further questions!

paales:
Interesting question @action-simon! For my own reasoning, i've tried writing out the situations:

We've got two situations:

1. The simple product is has visibility 'Not visible individually': The simple doesn't have it's own URL we only render the configurable URL. Canonical will be the configurable.
2. The simple product has visibility 'Catalog an/or Search': The simple does have it's own URL we switch the URL when an option is selected.

Situation 2 came from the desire to make the simple products inside configurables way more significant and reduce the significance of the configurable product. Let the configurable only function as a way to determine super attributes with variants.

The situation where the customer might be coming from: They weren't using configurables and had all simples visible in the catalog. They now want to bind all the simples together in a configurable product. By adding a configurable, the product page will now render with options. In this case the customer wants to keep the simple canonicals but in most other cases they dont.

A few questions arise that arise from me:

1. For both situations we currently replace all product information. Might it be an idea to actually limit the replacement of certain information when the simple doesn't have it's own URL? @carlocarels90 

2. When there is more than one super attribute, like color and size. You'd might want the color to be a different URL but don't want the size to affect the URL. We might have some information available, a seller can set one of the color variants to have a URL and when that color is selected it switches to that URL, but it doesn't switch when any of the other variants change?

3. Do we also want to reflect the selected options to the URL even when the simple products dont have their own URL? With a query param for example? This would make the URL sharable.

packages/magento-product-configurable/components/ConfigurableProductUrls/ConfigurableProductUrls.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { useConfigurableOptionsSelection } from '../../hooks/useConfigurableOptionsSelection'
+
+type ConfigurableProductUrlsProps = {
+  index?: number
+  product: Partial<ConfigurableOptionsFragment>
+}
+
+export function ConfigurableProductUrls(props: ConfigurableProductUrlsProps) {
+  const { product, index = 0 } = props
+  const router = useRouter()
```
paales, line 17: This will reduce retriggers of the useEffect because the replace never changes.

```suggestion
  const { replace } = useRouter()
```

packages/magento-product-configurable/components/ConfigurableProductUrls/ConfigurableProductUrls.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const variant = useConfigurableOptionsSelection({ url_key: product?.url_key, index }).configured
+    ?.configurable_product_options_selection?.variant
+  const urlKey = variant?.url_key ?? product?.url_key
+
+  const productLink = useProductLink({
+    url_key: urlKey,
+    __typename: variant?.__typename as ProductLinkProps['__typename'],
+  })
```
paales, line 25: I think it is better to not use useProductLink, but use productLink directly then it allows you to write something like this directly inside the useEffect:

```
const link = variant?.url_key ? productLink(variant) : productLink(product)
```

examples/magento-graphcms/pages/p/[url].tsx
```diff
@@ -73,23 +73,30 @@ function ProductPage(props: Props) {
...
+        ) : (
+          <JsonLd
+            item={{
+              '@context': 'https://schema.org',
+              ...jsonLdProduct(product),
+              ...jsonLdProductOffer(product),
+              ...jsonLdProductReview(product),
+            }}
+          />
+        )}
```
Comment (paales) [inline examples/magento-graphcms/pages/p/[url].tsx:93]: This needs to be moved to a plugin as well.

paales (changes requested):

paales (changes requested):

packages/magento-product-configurable/graphql/ConfiguredVariant.graphql
```diff
@@ -22,4 +13,10 @@ fragment ConfiguredVariant on SimpleProduct @injectable {
     }
     quantity
   }
+  url_rewrites {
+    url
+  }
+  ...ProductReviewsConfiguredVariant
+  ...ProductPageDescription
+  ...ProductPageMeta
```
paales, line 21: Might it be an idea to 'just' use the ProductPageItem fragment instead of the ConfiguredVariant and have everything work automatically?

paales (changes requested):

packages/magento-product/components/ProductName/ProductName.tsx
```diff
@@ -0,0 +1,11 @@
+import { ProductNameFragment } from './ProductName.gql'
+
+export type ProductNameProps = {
+  product: ProductNameFragment
```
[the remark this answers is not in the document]
carlocarels90: De `product` prop wordt in `ProductName.tsx` en in de plugin `ConfigurableProductName.tsx` gebruikt. Het voordeel van de product prop hier gebruik is dat je makkelijk ook andere data, zoals de `url_key` hier uit kan halen.

packages/magento-product-configurable/Config.graphqls
```diff
@@ -0,0 +1,52 @@
...
+  """
+  Use the meta of the configured variant
+  """
+  meta: Boolean
+
+  """
+  Use the shortDescription of the configured variant
+  """
+  shortDescription: Boolean
+}
```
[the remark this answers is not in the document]
carlocarels90: Do you mean we should combine **ProductDescription** + **ProductShortDescription** into `content`? I think that's a good idea

PR #1979, opened 2023-06-29
Reduce calls on success page which try to load an unused cart.

paales:


packages/magento-cart/hooks/useCartQuery.ts
```diff
@@ -14,30 +15,23 @@ import { useCurrentCartId } from './useCurrentCartId'
  */
 export function useCartQuery<Q, V extends { cartId: string; [index: string]: unknown }>(
   document: TypedDocumentNode<Q, V>,
-  options: QueryHookOptions<Q, Omit<V, 'cartId'>> & {
-    allowUrl?: boolean
-  } = {},
+  options?: QueryHookOptions<Q, Omit<V, 'cartId'>> & { returnPartialData: true },
+): QueryResult<DeepPartial<Q>, Omit<V, 'cartId'>>
```
mikekeehnen, line 19: Using `DeepPartial` might be exaggerated, since a lot of components are dependent on the data structure (Q) of the queries. Don't you think `Partial` will be good enough?

paales: An alternative course of action might be to remove the returnPartialData from the queries and try and get all the queries to be able to be resolved from the cache with `fetchPolicy: 'cache-only'`? We then need to figure out if queries actually return something if not, information needs to be queried earlier while the cart is still active?

PR #1990, opened 2023-07-12
fix: fix styling issues with algolia search

Giovanni-Schroevers:


examples/magento-graphcms/pages/search/index.tsx
```diff
@@ -93,11 +95,21 @@ function SearchResultPage(props: SearchResultProps) {
...
+            ))}
+          </CategorySearchResults>
         </Container>
 
+        <SearchDivider
+          sx={(theme) => ({
+            [theme.breakpoints.down('md')]: {
+              display: 'none',
+            },
+          })}
```
paales, line 110:
```suggestion
          sx={{ display: { xs: 'none', md: 'block' } }}
```

packages/algolia-search/plugins/AlgoliaCategorySearchResults.tsx
```diff
@@ -26,19 +27,19 @@ function CategoryHits() {
   )
 }
 
-function AlgoliaCategorySearchPlugin(props: PluginProps<SearchFormProps>) {
+function AlgoliaCategorySearchPlugin(props: PluginProps<CategorySearchResultsProps>) {
   const { Prev, ...rest } = props
```
paales, line 31:
```suggestion
function AlgoliaCategorySearchPlugin(props: PluginProps<CategorySearchResultsProps>) {
  const { Prev, children, ...rest } = props
```

packages/algolia-search/plugins/AlgoliaCategorySearchResults.tsx
```diff
@@ -26,19 +27,19 @@ function CategoryHits() {
...
   const searchIndex = useAlgoliaSearchIndexConfig('_categories')?.searchIndex
 
   if (!searchIndex) return <Prev {...rest} />
 
   return (
-    <>
-      <Prev {...rest} />
+    <Prev {...rest}>
+      {rest.children}
```
paales, line 38:
```suggestion
      {children}
```

packages/algolia-search/plugins/AlgoliaPageSearchResults.tsx
```diff
@@ -34,19 +34,19 @@ function PageHits() {
   )
 }
 
-function AlgoliaPageSearchPlugin(props: PluginProps<SearchFormProps>) {
+function AlgoliaPageSearchPlugin(props: PluginProps<CategorySearchResultsProps>) {
```
paales, line 37: Should we also rename the plugin to AlgoliaCategorySearchResultsPlugin?

paales (changes requested):

packages/algolia-search/plugins/AlgoliaPageSearchResults.tsx
Giovanni-Schroevers: This plugin is still responsible for the page index so I think the current name should be more applicable

paales:
@Giovanni-Schroevers I'm fine with merging this. Do you want to wait for the review of @mikekeehnen?

Giovanni-Schroevers:
@paales Seems fine to me as well but we aren't planning on doing another upgrade for a while so it is in no rush so if you would like to have @mikekeehnen take a look first we can wait on that

packages/algolia-search/components/SearchBox/SearchBox.tsx
```diff
@@ -51,7 +51,7 @@ export function SearchBox(props: SearchBoxProps) {
       inputRef={searchInputElement}
       onChange={debounceSearch}
       fullWidth
-      sx={{ mt: 1, mb: 1 }}
+      sx={[{ mt: 1, mb: 1 }, ...(Array.isArray(sx) ? sx : [sx])]}
```
mikekeehnen, line 54: Are the margins still relevant after using the `CategorySearchResults` component on the search page?

mikekeehnen (dismissed):

packages/magento-search/components/CategorySearchResult/CategorySearchResults.tsx
mikekeehnen, line 1: This component looks like it doesn't really have a purpose besides using it as anchor point for the plugin. Is this really necessary?

packages/magento-search/components/CategorySearchResult/CategorySearchResult.tsx
```diff
@@ -53,7 +52,6 @@ export function CategorySearchResult(props: CategorySearchResultProps) {
               text={breadcrumb?.category_name ?? ''}
               highlight={search}
             />
-            {' / '}
```
mikekeehnen, line 56: You may want to still have a `/` if the breadcrumbs show a path thats deeper then 1 level. Maybe only show a `/` if `length > 1` and `index > 0` ?

packages/magento-search/components/CategorySearchResult/CategorySearchResults.tsx
Giovanni-Schroevers: To me this seemed like a better solution than to abuse the `SearchForm` component for this. Using that component also gave some unintended styling changes to for example the search on QPlus

packages/algolia-search/components/SearchBox/SearchBox.tsx
Giovanni-Schroevers: We still want this margin if there are no search results so I would say yes

mikekeehnen (approved):

PR #1991, opened 2023-07-12
Feature/gcom 1120 | Improve defaultConfigurableOptionsSelection functionality 

JoshuaS98:
In this PR: 

**1. Improve defaultConfigurableOptionsSelection functionality**

_Situation: You land on a simple product page._ 

In the current main branch, we check all the configurableOptions of the linked configurable item. If a simple product does not have this option as a property on it's product branch, the option can not be configured on the simple product page. To fix this, you had to query that option in your productPage query to make it a property on the simple product. This however meant that you had to add a property to your query every time a site owner made a new configurable option. This is not dynamic nor scalable. 

To fix this:
We check the variants of a configurable option and match the variant SKU's with the SKU of the current simple product. When we know which variant should be preselected, we get the uid(s) from the variant's attribute(s) and inject them into the selectedOptions array. 

**2. Match the CORRECT related upsells with the CORRECT product**
In the current main branch, the first item of the relatedUpsells array is being merged into the product object. But in case of a simple product, magento not only gives the simple product back, but also optional configurable products that are linked to it. Merging relatedUpsells[0] with the simple product will result in merging the wrong upsells with the simple product.

**Test url** 
_Please replace localhost for Vercel URL._ 
Simple product: localhost:3000/p/waverly-size-4-6y-gc-564-sock-6y 
Concurrent configurable product: localhost:3000/p/gc-fruity-sock

**Issues:** 
1. If there is more than 1 configurable matched to a simple product, the simple product will only render the options of the first configurable returned in the productPage array.  This can cause potential issues, for instance, when someone is on the page of the second configurable linked to the simple, presses on an option, gets redirects to the simple and then suddenly some options that were unique on the second configurable are now longer showing as options. I did not validate this theory.

### DIFF
--- a/packages/magento-product-configurable/utils/defaultConfigurableOptionsSelection.ts
+++ b/packages/magento-product-configurable/utils/defaultConfigurableOptionsSelection.ts
@@ -24,9 +24,9 @@ export function defaultConfigurableOptionsSelection<Q extends BaseQuery = BaseQu
 
   const configurable = findByTypename(query?.products?.items, 'ConfigurableProduct')
   const variants = configurable?.variants
-  const simple = findByTypename(query?.products?.items, 'SimpleProduct')
+  const simple = requested
 
-  if (!configurable?.url_key || !simple) return { ...query, products: { items: [requested] } }
+  if (!configurable?.url_key) return { ...query, products: { items: [requested] } }
 
   const selectedOptions: string[] = []
 
--- a/packages/magento-product-configurable/utils/defaultConfigurableOptionsSelection.ts
+++ b/packages/magento-product-configurable/utils/defaultConfigurableOptionsSelection.ts
@@ -25,7 +25,6 @@ export function defaultConfigurableOptionsSelection<Q extends BaseQuery = BaseQu
   const configurable = findByTypename(query?.products?.items, 'ConfigurableProduct')
   const variants = configurable?.variants
   const simple = findByTypename(query?.products?.items, 'SimpleProduct')
-  const simpleSku = simple?.sku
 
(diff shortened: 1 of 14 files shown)


packages/magento-product-configurable/utils/defaultConfigurableOptionsSelection.ts
```diff
@@ -24,41 +23,39 @@ export function defaultConfigurableOptionsSelection<Q extends BaseQuery = BaseQu
...
+   * variant by comparing the sku of the current page with the available variants. => We then check
+   * the attributes of the variant and set the selectedOptions accordingly. => We want to always
+   * return the configurable item to a simple item. If we don't do this, the productpage will
+   * break.
+   *
+   * https://hoproj.atlassian.net/browse/GCOM-1120
+   */
+  variants?.forEach((v) => {
+    const vSku = v?.product?.sku
+    if (vSku === simpleSku) {
```
paales, line 52:
```suggestion
    if (v?.product?.uid === simple.uid) {
```

paales (changes requested):
I've tried moving away from the variants field as this could negatively impact performance. However, it seems that we are 'stuck' with it in this case.

paales:
> This can cause potential issues, for instance, when someone is on the page of the second configurable linked to the simple, presses on an option, gets redirects to the simple and then suddenly some options that were unique on the second configurable are now longer showing as options.

We currently never redirect a user and in the upcoming [ConfigurableProductUrl PR](https://github.com/graphcommerce-org/graphcommerce/pull/1978) only the URL is changed in the browser, but the page isn't actually reloaded.

It is kinda odd if a user will receive multiple configurables for the simple URL.

PR #1992, opened 2023-07-12
feat: export props from ProductPagePrice

Giovanni-Schroevers:


paales:
For future reference, if you're running into a situation like this and you need the props of a component to create a plugin for example, you can also do something like:

```ts
type ProductListPriceProps = React.ComponentProps<typeof ProductListPrice>
```

PR #1997, opened 2023-07-14
Feature/gcom 1125

Jessevdpoel:


examples/magento-graphcms/pages/checkout/success.tsx
```diff
@@ -56,7 +56,13 @@ function OrderSuccessPage() {
...
             </LayoutTitle>
             <CartSummary />
-            <CartItemSummary />
+            <CartItemSummary
+              sx={{
+                '& .Scroller-root.grid': {
+                  display: 'flex',
+                },
+              }}
+            />
```
paales, line 65: Shall we move this to inside the CartItemSummary, because I think it is a problem for all shops

packages/magento-cart/components/CartItemSummary/CartItemSummary.tsx
```diff
@@ -54,6 +54,9 @@ export function CartItemSummary(props: OrderSummaryProps) {
             theme.shape.borderRadius * 3,
             theme.breakpoints.values,
           ),
+          '& .Scroller-root.grid': {
+            display: 'flex',
+          },
```
paales, line 59: Shouldn't we configure this on the `<Scroller/>` directly?

paales (changes requested):

PR #2004, opened 2023-07-21
Feature/localstorage snackbar

bramvanderholst:


packages/next-ui/Snackbar/MessageSnackbar.tsx
```diff
@@ -1,4 +1,36 @@
...
+  const { storageType, id, onClose, ...rest } = props
+  const messageId = `MessageSnackBar_${id}`
+
+  if (storageType && !id && process.env.NODE_ENV !== 'production') {
+    console.error('MessageSnackbar with a storageType set must has an id.')
+  }
+
+  if (storageType && id && globalThis[storageType].getItem(messageId)) {
+    return null
+  }
```
paales, line 23:
```suggestion
  if (storageType && id && globalThis[storageType]?.getItem(messageId)) {
    return null
  }
```

paales (changes requested):
Can you provide a way to actually test this?

bramvanderholst:
@paales Can be tested here: https://graphcommerce-git-feature-localstorage-snackbar-graphcommerce.vercel.app/nl/test-page
I've added a message via `Row Column One` with the `Message` variant directly to the page. It can also be added to any or all pages using dynamic rows.

paales:
Depends on merge of https://github.com/graphcommerce-org/graphcommerce/pull/1938 so that this functionality can be implemented based on the dynamicrow functionality.

examples/magento-graphcms/components/GraphCMS/RowColumnOne/RowColumnOne.graphql
```diff
@@ -1,4 +1,6 @@
 fragment RowColumnOne on RowColumnOne {
+  id
+  rowColumnOneVariant
```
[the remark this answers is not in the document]
bramvanderholst: Don't think so, because; `Fields "variant" conflict because they return conflicting types "RowColumnOneVariants" and "RowProductVariants".`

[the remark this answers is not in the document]
paales: This doesn't have to be conflicting, you could also write:

`rowColumnOneVariant: variant`, so it is only renamed on the query side. What do you think?

bramvanderholst:
@paales The branch is up to date with canary and a working demo has been implemented using dynamic rows. 
The demo can be found here; https://graphcommerce-git-feature-localstorage-snackbar-graphcommerce.vercel.app/nl/test-page

Using dynamic rows the message was added to the page above and the page linked on that page.

Seems to work great when switching pages 🎉

packages/next-ui/Snackbar/MessageSnackbar.tsx
```diff
@@ -1,4 +1,36 @@
...
+      {...rest}
+      onClose={() => {
+        if (id && storageType) {
+          globalThis[storageType]?.setItem(messageId, `${Date.now()}`)
+        }
+        onClose?.()
+      }}
+    />
+  )
+}
```
paales, line 36: Rename to something like DismissableSnackbar and restore the old MessageSnackbar or move all logic to VariantMessage?

packages/next-ui/Snackbar/MessageSnackbarImpl.tsx
```diff
@@ -14,11 +14,12 @@ import React, { useEffect, useState } from 'react'
 import { IconSvg } from '../IconSvg'
 import { extendableComponent, breakpointVal } from '../Styles'
 import { iconClose, iconCheckmark, iconSadFace } from '../icons'
+import iconInfo from '../icons/info.svg'
 
 type Size = 'normal' | 'wide'
 type Variant = 'contained' | 'pill'
 
-export type MessageSnackbarProps = Omit<
+export type MessageSnackbarImplProps = Omit<
```
paales, line 22: The idea was that this is just an implementation detail, so the user didn't need to know

packages/next-ui/Row/ColumnOne/variant/VariantMessage.tsx
```diff
@@ -0,0 +1,26 @@
...
+        {...rest}
+        sx={[
+          {
+            pointerEvents: 'none',
+            '& > *': { pointerEvents: 'auto' },
+          },
+          ...(Array.isArray(sx) ? sx : [sx]),
+        ]}
+      />
+    </Portal>
```
paales, line 24: I would assume the Snackbar is already rendered in a Portal? Also the Portal component should be moved to the Impl side if this is really necessary.

examples/magento-graphcms/components/GraphCMS/RowColumnOne/RowColumnOne.tsx
```diff
@@ -1,17 +1,35 @@
...
+
+  if (!rowColumnOneVariant) return <Default {...RowColumnOneProps} />
+
+  const RenderType =
+    mergedRenderer?.[rowColumnOneVariant] ??
+    (() => {
+      if (process.env.NODE_ENV !== 'production')
+        return <>renderer for {rowColumnOneVariant} not found</>
+      return null
+    })
```
paales, line 32: Something like this?

```suggestion
  const { renderer, rowColumnOneVariant = 'Default', ...RowColumnOneProps } = props
  const mergedRenderer = { ...defaultRenderer, ...renderer } as VariantRenderer

  const RenderType = mergedRenderer?.[rowColumnOneVariant]
  if (!RenderType) return process.env.NODE_ENV !== 'production' ? <>renderer for {rowColumnOneVariant} not found</> : null
```

packages/next-ui/Theme/MuiSnackbar.ts
```diff
@@ -22,8 +22,10 @@ export const MuiSnackbar: SnackbarVariants = [
           },
         },
       '&.MuiSnackbar-anchorOriginBottomCenter': {
-        left: 0,
-        right: 0,
+        [theme.breakpoints.up('md')]: {
+          left: theme.page.horizontal,
+          right: theme.page.horizontal,
+        },
```
paales, line 28: We always lived without this, still necessary?

paales (changes requested):

packages/next-ui/Theme/MuiSnackbar.ts
bramvanderholst: Screenshot 2023-11-15 at 16 04 41
Without this, snackbars with content >= viewport width will stick to the sides of the viewport. Looks a bit odd.

examples/magento-graphcms/components/GraphCMS/RowColumnOne/RowColumnOne.tsx
bramvanderholst: `rowColumnOneVariant` can be `null`, so setting a default value like this won't work. However, I do like setting `rowColumnOneVariant` to `'Default'` more than an early return. I've updated the PR to use that instead.

packages/next-ui/Snackbar/MessageSnackbarImpl.tsx
```diff
@@ -54,8 +60,9 @@ export default function MessageSnackbarImpl(props: MessageSnackbarProps) {
     sticky,
     children,
     onClose,
-    severity = 'info',
+    severity = 'success',
```
paales, line 63: Maybe we should change it to 'info', and check usage sights and add 'success' there? Better to have an info icon (the one you added) by default I think :)

packages/next-ui/Snackbar/MessageSnackbarImpl.tsx
```diff
@@ -34,6 +35,11 @@ type OwnerState = {
   size?: Size
   severity?: 'success' | 'info' | 'warning' | 'error'
   variant?: Variant
+  /**
+   * Dialogs are dismissed on page interaction, whilst popups remain open on page interaction and
+   * can only be closed using the close button or escape key.
+   */
+  type?: 'dialog' | 'popup'
```
paales, line 42:
```suggestion
  type?: 'dialog' | 'modal'
```

packages/next-ui/Snackbar/DismissibleSnackbar.tsx
```diff
@@ -0,0 +1,31 @@
+import { MessageSnackbar } from './MessageSnackbar'
+import { MessageSnackbarProps } from './MessageSnackbarImpl'
+
+export type DismissibleSnackbarProps = MessageSnackbarProps & {
+  id?: string
```
paales, line 5:
```suggestion
  id: string
```

packages/next-ui/Row/ColumnOne/variant/VariantMessage.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { DismissibleSnackbar, DismissibleSnackbarProps } from '../../../Snackbar/DismissibleSnackbar'
+
+export type VariantMessageProps = DismissibleSnackbarProps
+
+export function VariantMessage(props: VariantMessageProps) {
+  const { sx, ...rest } = props
+
+  return (
+    <DismissibleSnackbar
+      open
```
paales, line 11:
```suggestion
```

packages/next-ui/Row/ColumnOne/variant/VariantMessage.tsx
```diff
@@ -0,0 +1,24 @@
...
+      variant='pill'
+      severity='info'
+      {...rest}
+      sx={[
+        {
+          pointerEvents: 'none',
+          '& > *': { pointerEvents: 'auto' },
+        },
+        ...(Array.isArray(sx) ? sx : [sx]),
+      ]}
```
paales, line 21:
```suggestion
      sx={sx}
```

packages/next-ui/Snackbar/DismissibleSnackbar.tsx
```diff
@@ -0,0 +1,31 @@
...
+  storageType?: 'localStorage' | 'sessionStorage'
+}
+export function DismissibleSnackbar(props: DismissibleSnackbarProps) {
+  const { storageType = 'localStorage', id, onClose, ...rest } = props
+  const messageId = `MessageSnackBar_${id}`
```
paales, line 10:
```suggestion
  const messageId = `MessageSnackBar_${id}`
  const [open, setOpen] = useState(false)

  // since we're syncing with an external system: useEffect
  useEffect(() => {
     setOpen(globalThis[storageType]?.getItem(messageId))
  }, [messageId])
```

packages/next-ui/Snackbar/DismissibleSnackbar.tsx
```diff
@@ -0,0 +1,31 @@
...
+    console.error('MessageSnackbar with a storageType set must has an id.')
+  }
+
+  if (storageType && id && globalThis[storageType]?.getItem(messageId)) {
+    return null
+  }
+
+  return (
+    <MessageSnackbar
+      {...rest}
```
paales, line 22:
```suggestion
      {...rest}
      open={open}
```

paales (changes requested):

PR #2005, opened 2023-07-27
Fix/conditional footer components

LeanderMatse:
If, for example, storeSwitcher is not passed, the corresponding container shouldn’t be rendered either. This prevents unnecessary HTML.

—

Gaps could be created with the current grid-template when removing an item. The grid could be rewritten to respond to adding or removing items, but that seems a bit to complex for a demo template. Developers can also easily re-assign a component to another grid-area.

### DIFF
--- a/.changeset/yellow-jobs-compare.md
+++ b/.changeset/yellow-jobs-compare.md
@@ -1,0 +1,5 @@
+---
+'@graphcommerce/next-ui': patch
+---
+
+Footer components should be rendered conditionally
--- a/.changeset/yellow-jobs-compare.md
+++ b/.changeset/yellow-jobs-compare.md
@@ -2,4 +2,4 @@
 '@graphcommerce/next-ui': patch
 ---
 
-Footer components should be rendered conditionally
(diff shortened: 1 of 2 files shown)


paales:
Hi @LeanderMatse! Thanks for the PR!

> Gaps could be created with the current grid-template when removing an item. The grid could be rewritten to respond to adding or removing items, but that seems a bit to complex for a demo template. Developers can also easily re-assign a component to another grid-area.

The deciding factor we've got here:
- If the rendering changes based on _data_ from the Magento backend (so if there is only one store for example), GraphCommerce should handle this properly
- If the rendering changes based on the customizations the developer does, we expect them to change the styles themselves.

.changeset/yellow-jobs-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphcommerce/next-ui': patch
+---
+
+Footer components should be rendered conditionally
```
paales, line 5:
```suggestion
Footer's grid-area's will only be rendered when the props are passed.
```

paales:
Thanks for the PR, merged!

PR #2006, opened 2023-07-28
Billing address was overwritten (GCOM-1137)

Jessevdpoel:


packages/magento-cart-shipping-address/components/CustomerAddressForm/CustomerAddressForm.tsx
```diff
@@ -33,8 +33,12 @@ export function CustomerAddressForm(props: CustomerAddressListProps) {
 
   const { data: cartQuery } = useCartQuery(GetAddressesDocument)
 
+  const defaultBillingAddress = customerAddresses.data?.customer?.addresses?.find(
+    (a) => a?.default_billing,
+  )
+
   const shippingAddress = cartQuery?.cart?.shipping_addresses?.[0]
-  const billingAddress = cartQuery?.cart?.billing_address
+  const billingAddress = defaultBillingAddress || cartQuery?.cart?.billing_address
```
paales, line 41: When we change this behavior, do we still get a valid billing address on the cart? Or will the billing address be empty on the cart / thus disabling the placeOrder?

paales (changes requested):

PR #2007, opened 2023-07-28
Crosssells improvements (GCOM-1138)

JoshuaS98:
[In this ticket](https://hoproj.atlassian.net/browse/GCOM-1138) implemented a more desirable behavior of the related products, inspired by wishes from [KOM-442](https://hoproj.atlassian.net/browse/KOM-442)

first check: @bramvanderholst 
second check: @paales 

### Most important changes:

**Magento Product**
- Crosssells should remain visible after add to cart. Do not pop away.
- Checkmark icon after add to cart added. showSuccess prop based on snack bar.
- Redirect default behavior off
- Correct product name to snackbar message after add crosssell product.

- Configure whether we want to redirect?
- Configure whether we want to filter out products that are in your cart?

**React hook form**
- add onComplete to RHF 
- add SubmittedVariables to RHF 

**Test**
/p/fuzzy-sticky-things-gc-691-sock

[KOM-442]: https://hoproj.atlassian.net/browse/KOM-442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/examples/magento-graphcms/pages/checkout/added.tsx
+++ b/examples/magento-graphcms/pages/checkout/added.tsx
@@ -33,10 +33,9 @@ function CheckoutAdded() {
   const { sku } = router.query
   const lastItem = items.find((item) => item.product.sku === sku)
 
-  const hideCrossSellItemsInCart =
-    false || Boolean(import.meta.graphCommerce.hideCrossSellItemsInCart)
+  const hideCrossSellItemsInCart = Boolean(import.meta.graphCommerce.hideCrossSellItemsInCart)
 
-  const redirectCrossSellItems = true || Boolean(import.meta.graphCommerce.redirectCrossSellItems)
+  const redirectCrossSellItems = Boolean(import.meta.graphCommerce.redirectCrossSellItems)
 
   const crosssels = useQuery(CrosssellsDocument, {
     variables: { pageSize: 1, filters: { sku: { eq: lastItem?.product.sku } } },
@@ -60,7 +59,6 @@ function CheckoutAdded() {
       items,
     ],
   )
-
   return (
     <>
       <PageMeta title={i18n._(/* i18n */ 'Cart')} metaRobots={['noindex']} />
--- a/examples/magento-graphcms/pages/checkout/added.tsx
+++ b/examples/magento-graphcms/pages/checkout/added.tsx
@@ -146,7 +146,7 @@ function CheckoutAdded() {
               <Trans id='Complete your purchase' />
             </Typography>
           </Container>
-          <AddProductsToCartForm disableSnackbar redirect={false}>
+          <AddProductsToCartForm disableSuccessSnackbar redirect={false}>
             <ItemScroller
               sx={(theme) => ({
                 width: 'auto',
--- a/examples/magento-graphcms/pages/checkout/added.tsx
+++ b/examples/magento-graphcms/pages/checkout/added.tsx
@@ -32,7 +32,11 @@ function CheckoutAdded() {
   const router = useRouter()
   const { sku } = router.query
   const lastItem = items.find((item) => item.product.sku === sku)
-  const redirectCrossSellItems = `${import.meta.graphCommerce.redirectCrossSellItems}`
+
+  const hideCrossSellItemsInCart =
+    false || Boolean(import.meta.graphCommerce.hideCrossSellItemsInCart)
+
+  const redirectCrossSellItems = true || Boolean(import.meta.graphCommerce.redirectCrossSellItems)
 
   const crosssels = useQuery(CrosssellsDocument, {
     variables: { pageSize: 1, filters: { sku: { eq: lastItem?.product.sku } } },
@@ -44,8 +48,17 @@ function CheckoutAdded() {
       filterNonNullableKeys(
         crosssels.data?.products?.items?.[0]?.crosssell_products ??
           crosssels.previousData?.products?.items?.[0]?.crosssell_products,
-      ).filter((item) => item.stock_status === 'IN_STOCK'),
-    [crosssels.data?.products?.items, crosssels.previousData?.products?.items],
+      ).filter(
+        (item) =>
+          (items.every((i) => i.product.sku !== item.sku) && hideCrossSellItemsInCart) ||
+          (item.stock_status === 'IN_STOCK' && !hideCrossSellItemsInCart),
+      ),
+    [
+      crosssels.data?.products?.items,
+      crosssels.previousData?.products?.items,
+      hideCrossSellItemsInCart,
+      items,
+    ],
   )
 
   return (
@@ -147,7 +160,10 @@ function CheckoutAdded() {
               <Trans id='Complete your purchase' />
             </Typography>
           </Container>
-          <AddProductsToCartForm disableSuccessSnackbar redirect={redirectCrossSellItems}>
+          <AddProductsToCartForm
+            disableSuccessSnackbar
+            redirect={redirectCrossSellItems ? 'added' : false}
+          >
             <ItemScroller
               sx={(theme) => ({
                 width: 'auto',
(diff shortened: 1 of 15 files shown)


examples/magento-graphcms/pages/checkout/added.tsx
```diff
@@ -43,11 +43,8 @@ function CheckoutAdded() {
       filterNonNullableKeys(
         crosssels.data?.products?.items?.[0]?.crosssell_products ??
           crosssels.previousData?.products?.items?.[0]?.crosssell_products,
-      ).filter(
-        (item) =>
-          items.every((i) => i.product.sku !== item.sku) && item.stock_status === 'IN_STOCK',
-      ),
-    [crosssels.data?.products?.items, crosssels.previousData?.products?.items, items],
+      ).filter((item) => item.stock_status === 'IN_STOCK'),
```
paales, line 46: We're removing the check of the SKU here, it seems to be fine, but can't guess what this actually changes? What changed here?

packages/magento-product/components/AddProductsToCart/useAddProductsToCartAction.ts
```diff
@@ -33,6 +37,31 @@ export function useAddProductsToCartAction(
...
+  const [showSuccess, setShowSuccess] = useState<boolean>(false)
+
+  const userErrors = toUserErrors(data)
+
+  const submitSuccesful =
+    !formState.isSubmitting &&
+    formState.isSubmitSuccessful &&
+    !error?.message &&
+    !userErrors.length &&
+    !redirect
```
paales, line 49: Do we want to only enable this behavior when there is `!redirect` or do we always want to enable this behavior?

packages/react-hook-form/src/useFormGql.tsx
```diff
@@ -28,7 +28,9 @@ export type UseFormGqlMethods<Q, V extends FieldValues> = Omit<
   UseGqlDocumentHandler<V>,
   'encode' | 'type'
 > &
-  Pick<UseFormReturn<V>, 'handleSubmit'> & { data?: Q | null; error?: ApolloError }
+  Pick<UseFormReturn<V>, 'handleSubmit'> & { data?: Q | null; error?: ApolloError } & {
+    submittedVariables?: V
+  }
```
paales, line 33:
```suggestion
  Pick<UseFormReturn<V>, 'handleSubmit'> & { data?: Q | null; error?: ApolloError; submittedVariables?: V }
```

packages/react-hook-form/src/useFormGql.tsx
```diff
@@ -76,6 +80,7 @@ export function useFormGql<Q, V extends FieldValues>(
       }
       // if (variables === false) onInvalid?.(formValues, event)
 
+      submittedVariables.current = variables
```
paales, line 83: Should we also empty the value when a user resubmits? I would expect the value to be set to undefined on line 72?

packages/magento-product/components/AddProductsToCart/AddProductsToCartSnackbar.tsx
```diff
@@ -35,8 +36,20 @@ export function AddProductsToCartSnackbar(props: AddProductsToCartSnackbarProps)
...
+    productSkus.push(item.sku)
+  })
+
+  const productsAdded = items
+    .filter((item) => productSkus.includes(item.product.sku))
+    .map((product) => product.product.name)
+
   const showErrorSnackbar = userErrors.length > 0
 
+  if (disableSnackbar) return null
```
paales, line 51: Should we rename this disableSuccess so the user still sees an error message when the product couldn't be added and only prevent the success message?

packages/magento-product/components/AddProductsToCart/AddProductsToCartSnackbar.tsx
```diff
@@ -35,8 +36,20 @@ export function AddProductsToCartSnackbar(props: AddProductsToCartSnackbarProps)
...
 
+  const productSkus: (string | undefined | null)[] = []
+
+  submittedVariables?.cartItems.forEach((item) => {
+    productSkus.push(item.sku)
+  })
+
+  const productsAdded = items
+    .filter((item) => productSkus.includes(item.product.sku))
+    .map((product) => product.product.name)
```
paales, line 47: I've got a feeling we can make this easier by not creating a productSkus directly but by using 'find' on the submittedVariables?.cartItems?

packages/magento-product/components/AddProductsToCart/useAddProductsToCartAction.ts
```diff
@@ -33,6 +37,31 @@ export function useAddProductsToCartAction(
...
+
+  useEffect(() => {
+    if (submitSuccesful && submittedVariables?.cartItems.find((item) => item.sku === sku)) {
+      setShowSuccess(true)
+    }
+  }, [sku, submitSuccesful, submittedVariables?.cartItems])
```
paales, line 55: We might be able to do this without useEffect: https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

paales (changes requested):

paales:
Demo omgeving: https://graphcommerce-git-feature-gcom-1138-graphcommerce.vercel.app/nl/p/free-candy-printable-birthday-postcard-gc-20-bdayprint

packages/magento-product/components/AddProductsToCart/AddProductsToCartForm.tsx
```diff
@@ -8,7 +8,7 @@ import {
 import { ExtendableComponent } from '@graphcommerce/next-ui'
 import { Box, SxProps, Theme, useThemeProps } from '@mui/material'
 import { useRouter } from 'next/router'
-import { useMemo, useRef } from 'react'
+import { useEffect, useMemo, useRef } from 'react'
```
bramvanderholst, line 11: @JoshuaS98 This newly added import does not seem to be used.

packages/magento-product/components/AddProductsToCart/AddProductsToCartSnackbar.tsx
```diff
@@ -69,7 +76,7 @@ export function AddProductsToCartSnackbar(props: AddProductsToCartSnackbarProps)
           <Trans
             id='<0>{name}</0> has been added to your shopping cart!'
             components={{ 0: <strong /> }}
-            values={{ name: items[items.length - 1]?.product.name }}
+            values={{ name: productsAdded.join(', ') }}
```
bramvanderholst, line 79: @JoshuaS98 Minor optimisation, but maybe use [Intl.ListFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat) for this?

We should also probably render a different sentence based on it being one or more products.
For a single product "has been added" is fine. For multiple products it should be "have been added".

packages/magento-product/components/AddProductsToCart/useAddProductsToCartAction.ts
```diff
@@ -32,6 +36,25 @@ export function useAddProductsToCartAction(
...
+
+  if (showSuccess) {
+    setTimeout(() => {
+      setShowSuccess(false)
+    }, 2000)
```
bramvanderholst, line 55: We should probably cancel this timeout when someone adds the same product multiple times before the timeout ends. Otherwise the checkmark will disappear too fast for the subsequent times the product is added.

packages/magento-product/components/AddProductsToCart/AddProductsToCartSnackbar.tsx
paales: Maybe the simple solution would be to do something like: 'Products have been added to your shopping cart!'. It doesn't happen too often though.

bramvanderholst (approved):

PR #2010, opened 2023-08-01
event.preventDefault() caused sign in field to not be focusable

Jessevdpoel:


paales:
It seems it might not be enough, might it be an idea to expand the functionality to also include the last few lines from this?

https://github.com/graphcommerce-org/graphcommerce/blob/783c000cc4c33b5d6f576f7e8832d7dac6529f59/packages/framer-scroller/hooks/useScroller.ts#L75-L83

PR #2011, opened 2023-08-02
Feature/gallery thumbnails with hover

mikekeehnen:


packages/framer-scroller/components/ScrollerThumbnail.tsx
```diff
@@ -0,0 +1,99 @@
...
+        theme.palette.primary.main,
+        theme.palette.action.hoverOpacity,
+      )}`,
+      `inset 0 0 0 2px #ffffff00, 0 0 0 4px #ffffff00`,
+    ],
+  )
+
+  const classes = withState({ active })
+
+  const scrollIntoView = () => ref.current?.scrollIntoView({ block: 'nearest', inline: 'nearest', behavior: 'instant' })
```
paales, line 52:
```suggestion
  const scrollIntoView = () => ref.current?.scrollIntoView({ block: 'center', inline: 'nearest', behavior: 'center' })
```

Might this be better?

packages/framer-scroller/plugins/GalleryThumbnailNavigation.tsx
```diff
@@ -0,0 +1,15 @@
+import type { PluginProps } from '@graphcommerce/next-config'
+import type { SidebarGalleryProps } from '@graphcommerce/next-ui/FramerScroller/SidebarGallery'
+import { useCallback } from 'react'
+import { ScrollerThumbnails } from '../components/ScrollerThumbnails'
+
+export const component = 'SidebarGallery'
+export const exported = '@graphcommerce/next-ui'
```
paales, line 7: ~~The location of this plugin is wrong, next-ui is dependent on framer-scroller, so framer-scroller can not create a plugin on next-ui. Just directly use it in SidebarGallery?~~

Already implemented in SidebarGallery, so can be removed

packages/next-ui/Config.graphqls
```diff
@@ -0,0 +1,24 @@
...
+input SidebarGalleryConfig {
+  """
+  This variable will enable or disable the sidebar gallery thumbnails.
+  """
+  thumbnails: Boolean
+  """
+  This variable will allocate the position of the sidebar gallery thumbnails.
+  """
+  thumbnailsPosition: SidebarGalleryThumbnailPosition
+}
```
paales, line 24: Something like this?

```suggestion
extend input GraphCommerceConfig {
  sidebarGallery: SidebarGalleryConfig
}

"""
Enumeration of all possible positions for the sidebar gallery thumbnails.
"""
enum SidebarGalleryPaginationVariant {
  DOTS
  THUMBNAILS_BOTTOM
}

"""
SidebarGalleryConfig will contain all configuration values for the Sidebar Gallery component.
"""
input SidebarGalleryConfig {
  """
  Variant used for the pagination
  """
  paginationVariant: SidebarGalleryVariant
}
```

paales (changes requested):

packages/framer-scroller/components/ScrollerThumbnail.tsx
mikekeehnen: Arguably this could be better. Difference is that the gallery always moves when changing the viewed image. See new deployment link to try it out

PR #2020, opened 2023-08-11
Update docs for Enable Guest Checkout Login setting

SimonPrins:


SimonPrins:
Added documentation for the Enable Guest Checkout Login setting introduced in Magento 2.4.6-p1.
This setting needs to be set correctly, otherwise customers will be unable to log in on GraphCommerce.

PR #2026, opened 2023-08-16
Chore/gcom 1126: Updated documentation and example

mikekeehnen:


packages/algolia-search/Config.graphqls
```diff
@@ -69,4 +69,14 @@ extend input GraphCommerceConfig {
...
+
+  """
+  Configure your Algolia HyGraph index. This will create a new index for HyGraph data if it doesn't exist in Algolia.
+  """
+  algoliaHygraphIndex: String
+
+  """
+  Configure your Algolia HyGraph secret. This will be used to authenticate the HyGraph data.
+  """
+  hygraphSecret: String
```
paales, line 81: We've got Hygraph variables available here as well should we use one of those? https://github.com/graphcommerce-org/graphcommerce/blob/canary/packages/hygraph-cli/Config.graphqls

paales (changes requested):

PR #2029, opened 2023-08-18
Feature/support official price including tax(gcom 1175)

Jessevdpoel:


packages/magento-cart-items/Api/CartItem.graphql
```diff
@@ -24,6 +24,9 @@ fragment CartItem on CartItemInterface @injectable {
     price {
       ...Money
     }
+    price_including_tax {
+      ...Money
+    }
```
paales, line 29: price_including_tax is only available from Magento 2.4.5 onwards, but since 2.4.4 is [supported until April of 2025](https://experienceleague.adobe.com/docs/commerce-operations/release/planning/lifecycle-policy.html?lang=en) we want to keep supporting that version.

To build in this backwards compatibility we should add the schema here that defines these fields which  https://github.com/graphcommerce-org/graphcommerce/blob/canary/packages/magento-graphql/schema/Magento243to245.graphqls

Something like this:

```
extend type CartItemPrices {
    price_including_tax:  Money
}
```

packages/magento-cart-items/components/CartItemActionCard/CartItemActionCard.tsx
```diff
@@ -42,6 +42,15 @@ export function CartItemActionCard(props: CartItemActionCardProps) {
...
   const inclTaxes = useDisplayInclTax()
 
+  const price = () => {
+    if (inclTaxes) {
+      return prices?.price_including_tax
+        ? prices.price_including_tax.value
+        : (prices?.row_total_including_tax?.value ?? 0) / quantity
+    }
+    return prices?.price.value
+  }
```
paales, line 52: This can just be a variable, doesn't need to be a function.

paales (changes requested):

PR #2033, opened 2023-08-24
Changed upperbound of resolveVal to 2560 (GCOM-1171)

Jessevdpoel:


packages/next-ui/Styles/responsiveVal.tsx
```diff
@@ -6,7 +6,7 @@
 export function responsiveVal(
   min: number,
   max: number,
-  maxBreakpoint = 1280,
```
mikekeehnen, line 9: Kunnen we bij deze waarde geen `theme.breakpoints.values.xl` gebruiken?

paales: responsiveVal doesn't have access to the theme object, so that isn't possible without modifications.

mikekeehnen: It would prevent future bugs that look like this, but might be a bit too much for a simple fix.

paales: I can discuss this later how to properly solve it, but for now I'd like to k eep it at 2560

PR #2036, opened 2023-08-25
feat(GCOM-1194): Implement Sticky AddToCart Component on Product Page

carlocarels90:


examples/magento-graphcms/pages/p/[url].tsx
```diff
@@ -66,109 +68,110 @@ function ProductPage(props: Props) {
     relatedUpsells?.items?.find((item) => item?.uid === products?.items?.[0]?.uid),
   )
 
+  const cartButtonRef = useRef<HTMLButtonElement | null>(null)
+
```
Comment (mikekeehnen) [inline examples/magento-graphcms/pages/p/[url].tsx:72]: Ultimately we don't want to create a ref like this on the product page. It would be nice if we could create a plugin on de `AddProductToCartButton` and move the sticky-add-to-cart logic to that plugin. We could portal that component a little bit higher in the DOM, to prevent weird styling issues https://react.dev/reference/react-dom/createPortal

examples/magento-graphcms/pages/p/[url].tsx
```diff
@@ -66,109 +68,110 @@ function ProductPage(props: Props) {
...
 
-          <ProductSidebarDelivery product={product} />
+        <ProductSidebarDelivery product={product} />
 
-          <ProductPageAddToCartActionsRow product={product}>
-            <AddProductsToCartButton fullWidth product={product} />
-            <ProductWishlistChipDetail {...product} />
-          </ProductPageAddToCartActionsRow>
+        <ProductPageAddToCartActionsRow product={product}>
+          <AddProductsToCartButton forwardedRef={cartButtonRef} fullWidth product={product} />
```
Comment (mikekeehnen) [inline examples/magento-graphcms/pages/p/[url].tsx:163]: See my previous comment :)

packages/magento-product/components/ProductPageWrapper/ProductPageWrapper.tsx
mikekeehnen, line 1: This component seems unused?

packages/magento-product/components/StickyAddToCart/StickyAddToCart.tsx
```diff
@@ -0,0 +1,163 @@
...
+      hideBackButton
+      forceScrolled={isButtonInViewport}
+      size='responsive'
+      sx={(theme) => ({
+        width: { md: '60vw', lg: '65vw' },
+        inset: 0,
+        margin: 'auto',
+        transform: {
+          xs: 'none',
+          md: `translateY(calc(${theme.appShell.appBarHeightMd}  * .28))`,
```
mikekeehnen, line 63: Where does this calculation come from? Why do you multiply the `appBarHeight` by .28?

mikekeehnen:
Screenshot 2023-08-25 at 14 20 46

This is a bug and should be happening. If we did not scroll yet, we don't want the button to appear

packages/magento-product/components/StickyAddToCart/StickyAddToCart.tsx
```diff
@@ -0,0 +1,163 @@
...
+                '& .incl-vat': {
+                  opacity: 0.4,
+                },
+              }}
+            >
+              <ProductPagePrice product={product} />
+            </Typography>
+          </AddProductsToCartError>
+        </ProductPageAddToCartQuantityRow>
+        <AddProductsToCartButton color='primary' size='medium' product={product} />
```
mikekeehnen, line 159: Needs styling on smaller screens

packages/magento-product/components/StickyAddToCart/StickyAddToCart.tsx
```diff
@@ -0,0 +1,163 @@
...
+            alt={mainImage.label ?? ''}
+            layout='fixed'
+            width={60}
+            height={60}
+            src={mainImage.url}
+            sx={(theme) => ({
+              verticalAlign: 'middle',
+              [theme.breakpoints.down('md')]: {
+                width: 46,
+                height: 46,
```
mikekeehnen, line 120: You might want to use a responsiveVal here

packages/magento-product/plugins/StickyAddToCart/StickyProductPageAddToCartRow.tsx
```diff
@@ -0,0 +1,41 @@
+import type { IfConfig, ReactPlugin } from '@graphcommerce/next-config'
+import { Box } from '@mui/material'
+import { useEffect, useRef, useState } from 'react'
+import { createPortal } from 'react-dom'
+import { ProductPageAddToCartRow, StickyAddToCart } from '../../components'
+
+export const component = 'ProductPageAddToCartActionsRow'
+export const exported =
+  '@graphcommerce/magento-product/components/ProductPage/ProductPageAddToCartRow'
+export const ifConfig: IfConfig = 'enableStickyAddToCart'
```
paales, line 10: This plugin is self-referencing. So the plugin is targeting the current package. Shouldn't we make this a simple if statement somewhere? @carlocarels90  / @mikekeehnen.

mikekeehnen: We could, but we could also make it a separate package right @paales? This would make it a plug and play solution

carlocarels90: @mikekeehnen if we do this, we don't need the config file I guess? Because it will apply only when you add this package to your project?

PR #2046, opened 2023-09-04
Decrease the fontSize for the payment button in the checkout to ensure the text fits within the button (GCOM-1186)

carlocarels90:
| Before | After |
|--------|--------|
| localhost_3000_checkout_payment(iPhone 12 Pro) (1) | localhost_3000_checkout_payment(iPhone 12 Pro) (2) |

### DIFF
--- a/.changeset/real-garlics-battle.md
+++ b/.changeset/real-garlics-battle.md
@@ -1,0 +1,5 @@
+---
+'@graphcommerce/magento-cart-payment-method': patch
+---
+
+decrease the fontSize for the payment button in the checkout to ensure the text fits within the button
--- a/.changeset/real-garlics-battle.md
+++ b/.changeset/real-garlics-battle.md
@@ -2,4 +2,4 @@
 '@graphcommerce/magento-cart-payment-method': patch
 ---
 
-decrease the fontSize for the payment button in the checkout to ensure the text fits within the button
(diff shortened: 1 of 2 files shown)


paales:
@carlocarels90 Might it be a better idea to remove the payment method name on the button. It was a 'cute' idea, but making the text smaller doesn't look like a great idea?

carlocarels90:
> @carlocarels90 Might it be a better idea to remove the payment method name on the button. It was a 'cute' idea, but making the text smaller doesn't look like a great idea?

@paales I think you're right. Adding other payment methods with even longer names in the future won't work either. I've reverted the font sizes and removed the payment method from the button.

paales:
@carlocarels90 Can you rebase the branch with canary?

carlocarels90:
> @carlocarels90 Can you rebase the branch with canary?

@paales done! ✅

.changeset/real-garlics-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphcommerce/magento-cart-payment-method': patch
+---
+
+decrease the fontSize for the payment button in the checkout to ensure the text fits within the button
```
paales, line 5:
```suggestion
Remove the selected payment method title from the Place order button so that the text always fits
```

paales (approved):

PR #2053, opened 2023-09-11
Created FormLayout component to push form control futher to the pages

mikekeehnen:
This PR provides the users of GraphCommerce more control over the "default" GraphCommerce Forms. Most of the forms will be extendable, replacable and editable.

An example of how this can be used:
```ts
<ShippingMethodForm step={4} sx={(theme) => ({ mt: theme.spacings.lg })}>
   {({original, control}) => (
   <>
   {original}
   <TextFieldElement control={control} name='company' />
   </>
   )}
</ShippingMethodForm>
```


packages/next-ui/Form/FormLayout.tsx
```diff
@@ -0,0 +1,19 @@
...
+    props: {
+      original: React.ReactNode
+    } & T,
+  ) => React.ReactNode
+  sx?: SxProps<Theme>
+}
+
+export function FormLayout<T>(props: FormLayoutProps<T>) {
+  const { children, original, form, sx } = props
+  return <Box sx={sx}>{children ? children({ original, ...form }) : original}</Box>
```
bramvanderholst, line 18: Wouldn't it be better to use a fragment instead of a Box so we don't introduce unnecessary div's.

mikekeehnen: This probably is a bit much. It's wrapped in a fragment now 👍

paales (changes requested):
Lets discuss this. I'm not necessarily a big fan of the callback style in children to do this.

Might it be an idea idea to make sure that created forms that accept possible children as customizations to also create a FormProvider and allow children to use something like https://www.react-hook-form.com/api/useformcontext/

We've done something like that for the AddProductsToCartForm as well (although it is a custom context, but the idea is the same).

paales:
@mikekeehnen Close?

PR #2067, opened 2023-09-26
Implement URL validation to prevent empty values for the HandleRedirect query

carlocarels90:


packages/magento-store/utils/redirectOrNotFound.ts
```diff
@@ -67,16 +67,17 @@ export async function redirectOrNotFound(
       candidates.add(from.endsWith(suffix) ? from.slice(0, -suffix.length) : `${from}${suffix}`)
     })
 
-    const routePromises = [...candidates].map(
-      async (url) =>
-        (
+    const routePromises = [...candidates].map(async (url) => {
```
paales, line 70: Easier solution might be: 

```suggestion
    const routePromises = [...candidates].filter(Boolean).map(async (url) => {
```

paales (changes requested):

packages/magento-store/utils/redirectOrNotFound.ts
carlocarels90: @paales I've added your suggestion 👍

PR #2070, opened 2023-10-02
Category page rel=prev/next

mikekeehnen:


packages/magento-product/components/ProductListPagination/ProductListPagination.tsx
```diff
@@ -22,9 +22,16 @@ export function ProductListPagination({
...
+        let rel: 'prev' | 'next' | undefined
+
+        if (btnProps.page && page_info.current_page) {
+          rel = btnProps.page <= page_info.current_page ? 'prev' : 'next'
+        }
+        
         return (
           <Link
             {...btnProps}
+            rel={rel}
```
paales, line 34: Rel prev/next are not a signal to google anymore.

https://developers.google.com/search/blog/2011/09/pagination-with-relnext-and-relprev
> rel=prev/next is not an indexing signal anymore. Check out our up-to-date [guide to pagination](https://developers.google.com/search/docs/specialty/ecommerce/pagination-and-incremental-page-loading).

paales: I'm not totally opposed to implementing this as it's relatively minor, but it should be implemented as a `link` tag, not a rel on an `a`

> In the past, Google used <link rel="next" href="..."> and <link rel="prev" href="..."> to identify next page and previous page relationships. Google no longer uses these tags, although these links may still be used by other search engines.

paales:
As Bing and Google do not use this as ranking indicators, we do not think this is usefull to implement

PR #2078, opened 2023-10-06
[GCOM-1247] | next-config watch options

KMalkowski:
Added @graphcommerce packages in node_modules to nextjs watch options

### DIFF
--- a/packagesDev/next-config/src/withGraphCommerce.ts
+++ b/packagesDev/next-config/src/withGraphCommerce.ts
@@ -166,7 +166,11 @@ export function withGraphCommerce(nextConfig: NextConfig, cwd: string): NextConf
 
       config.watchOptions = {
         ...(config.watchOptions ?? {}),
-        ignored: ['**/.git/**', `**/node_modules/!(@graphcommerce)**`, '**/.next/**'],
+        ignored: new RegExp(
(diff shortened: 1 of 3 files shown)


packagesDev/next-config/src/withGraphCommerce.ts
```diff
@@ -161,6 +164,11 @@ export function withGraphCommerce(nextConfig: NextConfig, cwd: string): NextConf
         ],
       }
 
+      config.watchOptions = {
+        ...(config.watchOptions ?? {}),
+        ignored: ['**/.git/**', `**/node_modules/!(@graphcommerce)**`, '**/.next/**'],
+      }
```
paales, line 170: I think this should be correct code, can we try this out? https://github.com/graphcommerce-org/graphcommerce/blob/canary/packagesDev/next-config/src/withGraphCommerce.ts#L157-L162

paales (changes requested):
After applying the suggested change can also run `yarn packages` in the root once so that the corresponding js file is generated and commit that as well?

packagesDev/next-config/src/withGraphCommerce.ts
KMalkowski: Yep, this works 👌

PR #2085, opened 2023-10-11
Bug/gcom 1210 broken product image animation

Jessevdpoel:


packages/framer-scroller/components/ScrollerButton.tsx
```diff
@@ -27,9 +27,17 @@ export const ScrollerButton = m(
       ...buttonProps
     } = props
 
+    const [isScrolling, setIsScrolling] = useState(false)
```
paales, line 30:
```suggestion
     const isScrolling = useRef(false)
```

packages/framer-scroller/components/ScrollerButton.tsx
```diff
@@ -27,9 +27,17 @@ export const ScrollerButton = m(
...
+
     const { getSnapPosition, scroll } = useScrollerContext()
     const scrollTo = useScrollTo()
-    const handleClick = () => scrollTo(getSnapPosition(direction))
+    const handleClick = async () => {
+      if (!isScrolling) {
+        setIsScrolling(true)
+        await scrollTo(getSnapPosition(direction))
+        setIsScrolling(false)
+      }
```
paales, line 39:
```suggestion
      if (!isScrolling.current) {
        isScrolling.current = true
        await scrollTo(getSnapPosition(direction))
        isScrolling.current =  false
      }
```

paales (changes requested):

packages/framer-scroller/components/ScrollerButton.tsx
```diff
@@ -27,9 +27,16 @@ export const ScrollerButton = m(
       ...buttonProps
     } = props
 
+    const isScrolling = useRef(false)
     const { getSnapPosition, scroll } = useScrollerContext()
     const scrollTo = useScrollTo()
-    const handleClick = () => scrollTo(getSnapPosition(direction))
+    const handleClick = async () => {
+      if (!isScrolling.current) {
```
mikekeehnen, line 34: I think we already have a `scroll.animating`. You can use that variable to catch the "spammed" click

paales (changes requested):
See changes from @mikekeehnen

packages/framer-scroller/components/ScrollerButton.tsx
```diff
@@ -1,7 +1,7 @@
 import { useMotionValueValue } from '@graphcommerce/framer-utils'
 import { Fab, FabProps, styled, SxProps, Theme } from '@mui/material'
 import { m, useTransform } from 'framer-motion'
-import React from 'react'
+import React, { useRef } from 'react'
```
paales, line 4: Not used

paales (approved):
This looks great! Last thing, can you clean up unused useRef?

Can we also create a followup ticket that will allow animating to the _next_ item when clicked multiple times?

PR #2088, opened 2023-10-11
Bug/cannot add multiple bundle items to cart gcom 1193

Jessevdpoel:


paales:
To be replaced with different PR

PR #2089, opened 2023-10-12
Use 'cache-first' fetch mode for getting 'total_quantity' data (GCOM-1241)

carlocarels90:
We need to create a cache entry to read it on the next reload. This change helps us store the 'total_quantity' data in cache, so we can read it again when needed.

Now, when a guest adds products to their cart and refreshes the page, the cart will still show what's in it. It won't be empty.

**Current state (Empty cart after page refresh):**
https://app.birdeatsbug.com/FTOieYax2X1CIARxbnf6DipqxDh5rVo9koyxckWrBSSb

**New state (No empty cart after page refresh):**
https://app.birdeatsbug.com/KpE-ExYKlW7PYdwxnugfiO6j-Sougwyo1Eu2pC5-UU5b

### DIFF
--- a/.changeset/lazy-seahorses-rhyme.md
+++ b/.changeset/lazy-seahorses-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@graphcommerce/magento-cart': patch
+---
+
+Use 'cache-first' fetch mode for getting 'total_quantity' data.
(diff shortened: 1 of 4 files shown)


paales:
This will cause an additional query to be made when adding a product to the cart, we'd like to prevent this. Lets discuss.

PR #2091, opened 2023-10-13
Update orderDetails and styling (GCOM-1255)

StefanAngenent:
https://hoproj.atlassian.net/browse/GCOM-1255


packages/magento-customer/components/OrderDetails/OrderDetails.tsx
```diff
@@ -105,255 +43,230 @@ export function OrderDetails(props: OrderDetailsProps) {
...
                 {invoices && invoices?.length > 0 && (
-                  <Invoice>
+                  <Box
+                    className={classes.invoice}
+                    sx={(theme) => ({
+                      display: 'flex',
+                      justifyContent: 'flex-start',
+                      alignItems: 'center',
+                      color: theme.palette.primary.main,
+                    })}
```
paales, line 159: This currently has a color, but you can't click on it. Do we want to remove the color here and create an issue here?

packages/magento-customer/components/OrderDetails/OrderDetails.tsx
```diff
@@ -105,255 +43,230 @@ export function OrderDetails(props: OrderDetailsProps) {
...
+            <SectionContainer
+              variantLeft='h5'
+              labelLeft={<Trans id='Order status' />}
+              className={classes.orderDetailTitle}
+              sx={{ '& .SectionHeader-root': { marginTop: 0, paddingBottom: '4px' } }}
+            />
+            <Typography>
+              <Trans id='Ordered items' />:{' '}
+              {order_date && dateFormatter.format(new Date(order_date))}
+            </Typography>
```
paales, line 113: The SectionContainer header is called 'Order status', we're not displaying the order status here, is this something we want?

Also shouldn't the content be inside the SectionContainer?

packages/magento-customer/components/OrderDetails/OrderDetails.tsx
```diff
@@ -105,255 +43,230 @@ export function OrderDetails(props: OrderDetailsProps) {
...
+            </SectionContainer>
+          </Box>
+        </Box>
+      </SectionContainer>
 
-      <TotalsContainer>
-        <TotalsRow>
-          <div>
+      <Box
+        className={classes.totalsContainer}
```
paales, line 193: Do we want to move the OrderTotals to a separate component and move the OrderItems to in between, so that we have this order?

1. OrderOverview
2. OrderItems
3. OrderTotals

paales (changes requested):

packages/magento-customer/components/OrderDetails/OrderDetails.tsx
StefanAngenent: Yes, I've changed it around. It's more in line with the checkout now.

packages/magento-customer/components/OrderDetails/OrderDetails.tsx
StefanAngenent: Yes this should be inside the SectionContainer. 

In the old situatIon the order satus was the same, think it was misspeld, I changed it to the 'Order Date' now. The order status would be nice to implement here. I added it below the headertitle.

PR #2098, opened 2023-10-19
[GCOM-1266] Fix editAddressForm when addition is nullable

JoshuaS98:
- [GCOM Ticket](https://hoproj.atlassian.net/browse/GCOM-1266)
- [On request of Bouwprofi](https://hoproj.atlassian.net/browse/BPGC-215)

### Error
Magento gave an internal error when addition field was null or undefined. This was a magento bug but simply fixed by sending an empty string if addition is not given
image

### DIFF
--- a/packages/magento-customer/components/EditAddressForm/EditAddressForm.tsx
+++ b/packages/magento-customer/components/EditAddressForm/EditAddressForm.tsx
@@ -50,7 +50,7 @@ export function EditAddressForm(props: EditAddressFormProps) {
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
         houseNumber: address?.street?.[1],
-        addition: address?.street?.[2],
+        addition: address?.street?.[2] ?? '',
       },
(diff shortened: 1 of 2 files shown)


packages/magento-customer/components/EditAddressForm/EditAddressForm.tsx
```diff
@@ -50,7 +50,7 @@ export function EditAddressForm(props: EditAddressFormProps) {
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
         houseNumber: address?.street?.[1],
```
paales, line 52: This can also be empty in theory. While you're at it, can you try and remove the ts-ignore?

paales (changes requested):

packages/magento-customer/components/EditAddressForm/EditAddressForm.tsx
paales: I'm not completely sure, but if it is currently working I wouldn't change it.

PR #2100, opened 2023-10-23
Feature/GCOM-1108 | Property picker for Dynamic Rows (V2)

JoshuaS98:
- [Ticket](https://hoproj.atlassian.net/browse/GCOM-1108) 

### Log

Built a working fix (#37cb9a6 | feature/GCOM-1108. Tested for ConditionAnd and ConditionOr. These both work. A black friday text can be seen on the following urls, among others:

- http://localhost:3000/p/paying-hush-gc-923-sock
- http://localhost:3000/p/subtle-sympathy-gc-1087-sock

I also straightened out the styling with that of the Hygraph UI.

**Some thoughts:** 

1. ~~How do we determine the inputs for the property field? Right now it is created with a manual graphql query. Something with an introspection query?~~
2. ~~Can we then also filter the input fields for ConditionNumber and ConditionText? So for ConditionNumber, show only properties that expect an integer. And vice versa.~~

4. ~~How do we want to render the tags?~~
5. ~~The app should be hosted on a NextJS page or on an external application.~~
6. ~~A setup.tsx should be written for the app.~~
7. ~~Documentation should be written for installing the application~~
8. ~~Add changeset~~


packages/hygraph-app/pages/drpp-setup.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import { Page } from '..'
+
+export default function Setup() {
+  const appContainer = React.useRef(null)
```
mikekeehnen, line 5: Ref should be properly typed

packages/hygraph-app/pages/hygraph/property-picker.tsx
```diff
@@ -0,0 +1,105 @@
...
+      if (!a.label.includes('.') && !b.label.includes('.')) {
+        return a.label.localeCompare(b.label)
+      }
+      if (a.label.includes('.')) {
+        return 1
+      }
+      return -1
+    }),
+    number: [...numberOptions, { label: 'url', id: 'url' }],
+  }
```
mikekeehnen, line 49: You might want to memo this object

packages/hygraph-app/pages/hygraph/property-picker.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  React.useEffect(() => {
+    /**
+     * Some styling is being undone here to resolve conflicts between Hygraph App SDK and CssAndFramerMotionProvider.
+     */
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
mikekeehnen, line 56: You shouldn't ignore valid warnings

packages/hygraph-app/lib/objectifyGraphQLInterface.ts
```diff
@@ -0,0 +1,62 @@
...
+        [value?.name]: typeValue,
+      }
+    } else if (conditionType === typeValue) {
+      objectifiedInterface = {
+        ...objectifiedInterface,
+        [value?.name]: typeValue,
+      }
+    } else if (conditionType !== typeValue) {
+      // do nothing
+    }
```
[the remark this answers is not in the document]
mikekeehnen:
```
else if (conditionType !== typeValue) {
      // do nothing
    }
```
This statement doesn't do anything right now. You could remove it right?

packages/hygraph-app/lib/createRecursiveIntrospectionQuery.ts
```diff
@@ -0,0 +1,13 @@
+export const createRecursiveIntrospectionQuery = (type, depth) => {
```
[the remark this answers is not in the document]
mikekeehnen: At the places you are using this recursive function, you are still defining a depth. This means you do know the dept it is going to be

packages/hygraph-app/lib/createOptionsFromInterfaceObject.ts
```diff
@@ -0,0 +1,42 @@
+import { ProductProperty, ObjectType } from '../types'
+
+export const createOptionsFromInterfaceObject = (
+  obj: ObjectType,
```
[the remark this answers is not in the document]
mikekeehnen: You could just replace it and everything wil work the same

packages/hygraph-dynamic-rows-ui/lib/createRecursiveIntrospectionQuery.ts
```diff
@@ -0,0 +1,13 @@
+export const createRecursiveIntrospectionQuery = (type, depth) => {
```
mikekeehnen, line 1: You should type these props

packages/hygraph-dynamic-rows-ui/lib/fetchGraphQLInterface.ts
```diff
@@ -0,0 +1,14 @@
+import { gql } from '@apollo/client'
+import { createRecursiveIntrospectionQuery } from './createRecursiveIntrospectionQuery'
+
+export const fetchGraphQLInterface = (client) => {
```
mikekeehnen, line 4: You should type this prop

packages/hygraph-app/lib/createRecursiveIntrospectionQuery.ts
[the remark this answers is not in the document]
mikekeehnen: I don't see the need for a recursive function to generate graphql templates, since this makes the code less readable and nothing is automated by doing it this way. One pro of graphql is an easy to understand syntax, which you lose by doing it this way

mikekeehnen (approved):

packages/hygraph-dynamic-rows-ui/pages/property-picker.tsx
```diff
@@ -0,0 +1,107 @@
...
+export const getStaticProps = async () => {
+  const config = loadConfig(process.cwd())
+  const staticClient = new ApolloClient({
+    link: new HttpLink({
+      uri: config.magentoEndpoint,
+      fetch,
+    }),
+    cache: new InMemoryCache(),
+  })
+  const graphQLInterface = fetchGraphQLInterface(staticClient)
```
paales, line 100: This calls the demo backend, shouldn't this call the users' backend? Is this something we can configure somehow?

packagesDev/next-config/dist/generated/config.js
```diff
@@ -1,110 +1,77 @@
...
-        content: _zod.z.boolean().nullish(),
-        gallery: _zod.z.boolean().nullish(),
-        url: _zod.z.boolean().nullish()
+    return zod_1.z.object({
+        content: zod_1.z.boolean().nullish(),
+        gallery: zod_1.z.boolean().nullish(),
+        url: zod_1.z.boolean().nullish()
     });
 }
+exports.MagentoConfigurableVariantValuesSchema = MagentoConfigurableVariantValuesSchema;
```
paales, line 77: Always run `yarn codegen` in the example dir after using `yarn packages` in the root

packages/hygraph-dynamic-rows-ui/pages/_document.tsx
```diff
@@ -0,0 +1,13 @@
...
+  return (
+    <Html lang='en'>
+      <Head />
+      <body>
+        <Main />
+        <NextScript />
+      </body>
+    </Html>
+  )
+}
```
paales, line 13: _document isn't required.

paales (changes requested):
It seems the app isn't deployed yet, so can't be tested properly.

PR #2102, opened 2023-10-25
feat(GCOM-1245): autofocus email field in checkout

FrankHarland:


packages/magento-cart-email/EmailForm/EmailForm.tsx
```diff
@@ -69,6 +69,7 @@ const EmailFormBase = React.memo<EmailFormProps>((props) => {
           })}
           InputProps={{
             autoComplete: 'email',
+            autoFocus: true,
```
paales, line 72: Do we always want to autoFocus this? Maybe better to pass this by a prop, seems a configuration specific to the place where the component is used.

paales (changes requested):

paales:
Marked this as stale, not sure if we should proceed, you had some second thoughts?

FrankHarland:
> Marked this as stale, not sure if we should proceed, you had some second thoughts?

@paales Yes, while working on the a11y improvements I found multiple sources claiming autofocus to be possibly quite annoying for impaired users. In the newer GC versions we even have a seperate button offering the user a choice between signin and registration. If we autofocus the field we bypass this option and things can get confusing.

So I think it might be better to just scrap this PR.

PR #2107, opened 2023-11-03
feat(GCOM-1237): remember email adress when going from checkout to si…

FrankHarland:
…gnin//up form

### DIFF
--- a/packages/magento-customer/components/AccountSignInUpForm/AccountSignInUpForm.tsx
+++ b/packages/magento-customer/components/AccountSignInUpForm/AccountSignInUpForm.tsx
@@ -11,6 +11,7 @@ import { Trans } from '@lingui/react'
 // eslint-disable-next-line @typescript-eslint/no-restricted-imports
 import { Box, CircularProgress, Link, SxProps, TextField, Theme, Typography } from '@mui/material'
 import router from 'next/router'
+import { useEffect } from 'react'
 import { CustomerDocument, useFormIsEmailAvailable } from '../../hooks'
 import { useCustomerQuery } from '../../hooks/useCustomerQuery'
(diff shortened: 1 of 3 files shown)


packages/magento-customer/components/AccountSignInUpForm/AccountSignInUpForm.tsx
```diff
@@ -33,9 +34,18 @@ export function AccountSignInUpForm(props: AccountSignInUpFormProps) {
...
   const disableFields = formState.isSubmitting && !autoSubmitting
 
+  useEffect(() => {
+    const emailFromParams = router.query.email as string
+    if (!email && emailFromParams) {
+      setValue('email', emailFromParams)
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      trigger('email')
+    }
+  }, [email, setValue, trigger])
```
paales, line 47: Might be better to move emailFromParams out of the useEffect, you probably have a warning on the dependency array right now.

paales (changes requested):

packages/magento-customer/components/AccountSignInUpForm/AccountSignInUpForm.tsx
```diff
@@ -11,6 +11,7 @@ import { Trans } from '@lingui/react'
 // eslint-disable-next-line @typescript-eslint/no-restricted-imports
 import { Box, CircularProgress, Link, SxProps, TextField, Theme, Typography } from '@mui/material'
 import router from 'next/router'
```
paales, line 13: Not allowed, please use useRouter

FrankHarland: Changed it :)

PR #2110, opened 2023-11-07
Feature/new guestcheckout disabled gcom 1141

Jessevdpoel:


packages/magento-customer/components/AccountSignInUpForm/AccountSignInUpForm.tsx
```diff
@@ -1,27 +1,15 @@
...
-import { SignInForm } from '../SignInForm/SignInForm'
-import { SignUpForm } from '../SignUpForm/SignUpForm'
+import { AccountSignInUpIsEmailAvailableForm } from './AccountSignInUpIsEmailAvailableForm'
+import { AccountSignInUpToggleForm } from './AccountSignInUpToggleForm'
 
-export type AccountSignInUpFormProps = { sx?: SxProps<Theme> }
+// When the guestCheckout is disabled in Magento add loginMethod "TOGGLE" to your config to enable the toggle login / register flow which is mandatory.
+// Toggle flow also disables the isEmailAvailable call to the magento backend.
+
+// Default is set to IS_EMAIL_AVAILABLE
```
mikekeehnen, line 10: Isn't the guestCheckout value retrievable via the store config in Magento? If so, we should toggle this behaviour automatically, since wrong configuration could lead to faulty UI

packages/magento-cart-email/EmailForm/EmailForm.tsx
```diff
@@ -71,7 +71,8 @@ const EmailFormBase = React.memo<EmailFormProps>((props) => {
             autoComplete: 'email',
             endAdornment: (
               <WaitForQueries waitFor={isEmailAvailable}>
-                {isEmailAvailable.data?.isEmailAvailable && (
+                {(isEmailAvailable.data?.isEmailAvailable ||
```
mikekeehnen, line 74: Isn't `isEmailAvailable` always `true` when the guest checkout is disabled? That makes the `isEmailAvailable.data?.isEmailAvailable` redundant

mikekeehnen (changes requested):
This doesn't meet the design, the buttons are way to close to the fields.

packages/magento-customer/components/AccountSignInUpForm/AccountSignInUpForm.tsx
Jessevdpoel: No the guestCheckout value is not retrievable via the store config in Magento. If we set guestCheckout disabled in the config the IsEmailAvailable API will always return true.

PR #2117, opened 2023-11-20
Fixes for windows installation

kim-ep:
I've added some fixes to GraphCommerce source code that allowed to install it on Windows 11 without using WSL.

List of changes:
- escape spaces in packages.json scripts with double quotes instead of single quotes
- support glob pattern when using `rimraf` package (it's disabled by default for Windows)
- replace back slashes with forward slashes when POSIX-style path is implied

Unfortunately, changes related to back slashes are masked by code reformatting forced by Prettier.
In `findPlugins.ts` only one change is meaningful, in the line 68:
```
    .replace(dependency.replace(/\\/g, '/'), path)
```
In `generateInterceptors.ts` only one change is meaningful, in the line 215:
```
    .replace(/\\/g, '/')
```


paales:
@kim-ep I've tried adding a windows build to github actions, lets see if that works.

PR #2125, opened 2023-11-27
Bug/gcom 1258 overlay closes on esc

Jessevdpoel:


Jessevdpoel:
Changed 'keyup' to 'keydown' because MUI elements use a stopropagation in their keydown events. Because of this the event handler does not handle the keydown event of the overlay if there is an open MUI element. If the MUI element is closed  the keydown event of the overlay is called.

PR #2127, opened 2023-11-27
Googleanalytics: remove hard dependency on @graphcommerce/magento* packages (GCOM-1216)

Jessevdpoel:


packages/googletagmanager/package.json
```diff
@@ -23,5 +23,10 @@
     "next": "^14",
     "react": "^18.2.0",
     "react-dom": "^18.2.0"
+  },
+  "peerDependenciesMeta": {
+    "@graphcommerce/magento-product":{
+      "optional": true
+    }
```
paales, line 30: Not required here?

paales (changes requested):

PR #2128, opened 2023-11-27
Bug/gcom 1291 cannot delete selected region

Jessevdpoel:


packages/magento-customer/components/AddressMultiLine/AddressMultiLine.tsx
```diff
@@ -52,7 +52,7 @@ export function AddressMultiLine(props: AddressMultiLineProps) {
         {postcode} {city}
       </div>
       <div>
-        {regionName && `${regionName}, `}
+        {regionName && region?.region_id && `${regionName}, `}
```
paales, line 55: It is being fixed on the render side here, does this still mean that address submitted to Magento is wrong? Because that would of course need to be fixed as well, right?

paales (changes requested):

PR #2130, opened 2023-11-29
Feature/gcom 1221

FrankHarland:
Various a11y improvements.
(See this branch vs canary branch when using tab based navigation with voiceover enabled) (cmd+fn+f5, to toggle)


examples/magento-graphcms/components/Layout/LayoutFull.tsx
```diff
@@ -62,7 +62,7 @@ export function LayoutFull(props: LayoutFullProps) {
             )}
             <Fab
               href='/service'
-              aria-label={i18n._(/* i18n*/ `Account`)}
+              aria-label={i18n._(/* i18n*/ `CustomerService`)}
```
bramvanderholst, line 65:
```suggestion
              aria-label={i18n._(/* i18n*/ `Customer Service`)}
```

examples/magento-graphcms/components/Layout/LayoutNavigation.tsx
```diff
@@ -29,6 +29,7 @@ import { useRouter } from 'next/router'
 import { Footer } from './Footer'
 import { LayoutQuery } from './Layout.gql'
 import { Logo } from './Logo'
+import { global } from 'styled-jsx/css'
```
bramvanderholst, line 32: Unused import?

examples/magento-graphcms/components/Layout/LayoutNavigation.tsx
```diff
@@ -122,7 +123,15 @@ export function LayoutNavigation(props: LayoutNavigationProps) {
...
               ))}
 
-              <DesktopNavItem onClick={() => selection.set([menu?.items?.[0]?.uid || ''])}>
+              <DesktopNavItem
+                onClick={() => selection.set([menu?.items?.[0]?.uid || ''])}
+                onKeyUp={(evt) => {
+                  if (evt.key === 'Enter') {
+                    selection.set([menu?.items?.[0]?.uid || ''])
+                  }
+                }}
```
bramvanderholst, line 132:
```suggestion
                onKeyDown={(evt) => {
                  if(evt.key === ' ') {
                    evt.preventDefault()
                  }
                }}
                onKeyUp={(evt) => {
                  if (evt.key === 'Enter' || evt.key === ' ') {
                    selection.set([menu?.items?.[0]?.uid || ''])
                  }
                }}
```

examples/magento-graphcms/pages/checkout/added.tsx
```diff
@@ -24,6 +25,11 @@ type GetPageStaticProps = GetStaticProps<LayoutOverlayProps, Props>
 
 function CheckoutAdded() {
   const [addedItem, crossSellItems] = useCrosssellItems()
+  const a11yFocusRef = useRef<HTMLHeadingElement | null>(null)
+
+  useEffect(() => {
+    a11yFocusRef.current?.focus()
+  })
```
bramvanderholst, line 32: Not passing a dependency array will cause this to happen every render. Should this not happen only on "mount" (by adding an empty dependency array)?

packages/framer-scroller/components/ScrollerDots.tsx
```diff
@@ -7,8 +7,6 @@ import React from 'react'
 import { useScrollerContext } from '../hooks/useScrollerContext'
 import { ScrollerDot } from './ScrollerDot'
 
-const MotionBox = styled(m.div)({})
```
bramvanderholst, line 10: Import for styled can be removed

packages/next-ui/Navigation/components/NavigationFab.tsx
```diff
@@ -54,19 +59,20 @@ export function NavigationFab(props: NavigationFabProps) {
...
           [theme.breakpoints.down('md')]: {
             opacity: '1 !important',
             transform: 'none !important',
           },
-        })}
+        }}
         style={{ opacity }}
       >
         <Fab
+          tabIndex={!isMobile && !fabIsVisible ? -1 : 0}
```
bramvanderholst, line 71: Why does this behaviour need to be different on mobile & desktop?

packages/magento-search/components/SearchLink/SearchLink.tsx
```diff
@@ -38,13 +38,24 @@ export function SearchLink(props: SearchLinkProps) {
...
       <Link
         component='a'
         className={classes.root}
         underline='none'
         onClick={handleClick}
+        onKeyDown={(e) => {
+          if (e.key === 'Enter') {
+            handleKeyDown(e)
+          }
+        }}
```
mikekeehnen, line 57:
```suggestion
  const handleKeyDown: KeyboardEventHandler<HTMLAnchorElement> = (e) => {
    if (e.key === 'Enter') {
        e.preventDefault()
        return router.push(href)
    }
  }

  return (
    <>
      <Link
        component='a'
        className={classes.root}
        underline='none'
        onClick={handleClick}
        onKeyDown={handleKeyDown}
```

packages/next-ui/TextInputNumber/TextInputNumber.tsx
```diff
@@ -163,7 +163,7 @@ export function TextInputNumber(props: TextInputNumberProps) {
       }}
       inputProps={{
         ...inputProps,
-        'aria-label': i18n._(/* i18n */ 'Number'),
+        'aria-label': i18n._(/* i18n */ 'Add to cart quantity'),
```
paales, line 166: Please change on the usage side, since this field can be used for more fields

packages/magento-review/components/ProductReviewChip/ProductReviewChip.tsx
```diff
@@ -36,6 +36,7 @@ export function ProductReviewChip(props: ProductReviewChipProps) {
       icon={<IconSvg src={iconStar} sx={{ stroke: '#FFDA1C', fill: '#FFDA1C' }} />}
       color='default'
       size='medium'
+      aria-label='Review score'
```
paales, line 39: Translations

examples/magento-graphcms/pages/checkout/added.tsx
```diff
@@ -77,15 +83,15 @@ function CheckoutAdded() {
         )}
 
         <Box gridArea='children'>
-          <Box sx={{ typography: 'h6' }}>
+          <Box sx={{ typography: 'h6' }} tabIndex={-1} ref={a11yFocusRef}>
```
paales, line 86: Focussing on a `tabIndex={-1}`?

paales (changes requested):

examples/magento-graphcms/pages/checkout/added.tsx
FrankHarland: Yes, the H6 normally should not be focusable. (Thus -1) but when the user adds an item to his cart we do want to shift focus to the modal. (And in this case have the screenreader tell the user the action worked)

packages/next-ui/Navigation/components/NavigationFab.tsx
FrankHarland: Because the menu fab is initially hidden until the user scrolls a bit. So it would be better if it was only focusable when it's actually visible according to general WCAG. However in this case i think it would be better to just leave it as-is. As fiddling about with this might create performance issues And the button itself remains useable even when not yet scrolled.

Most of the time the user will be interacting with the page though, so I've removed this check and just left the menu button in the normal tab flow.

packages/next-ui/TextInputNumber/TextInputNumber.tsx
FrankHarland: Done!

paales (approved):

PR #2132, opened 2023-11-30
Fix/magento compare button hover

action-simon:
Fix compare button hover state on product page


paales:
Thanks for the PR!

This seems to be adding a boxShadow on any product grid item: https://graphcommerce-git-fork-action-simon-fix-ma-dfd38f-graphcommerce.vercel.app/nl/dames/business

PR #2139, opened 2023-12-05
Multiple dynamic rows (GCOM-1307)

JoshuaS98:
- Migration script for 7.1 
- enabled multiple rows per DR


paales:
@JoshuaS98 Please make sure older versions of GraphCommerce keep working as expected. Please put back the row and deprecate it.

PR #2147, opened 2023-12-12
fix: Added loading state to useFormGql

mikekeehnen:


paales:
I think this can already be achieved by: `formState.isSubmitting`

mikekeehnen:
> I think this can already be achieved by: `formState.isSubmitting`

`isSubmitting` gets triggered on form submission and validation, not on the actual mutation loading state. Might be good to separate those values like this.

paales:
@mikekeehnen Can PM me an example how this is used?

paales:
No apparent use cases communicated.

PR #2151, opened 2023-12-14
[GCOM-1066]: Added telephone to address overview

mikekeehnen:


packages/magento-customer/components/AccountAddress/AccountAddress.tsx
```diff
@@ -31,7 +31,7 @@ export function AccountAddress(props: AccountAddressProps) {
     >
       <Box className={classes.address} sx={{ '& > span': { display: 'block' } }}>
         <AddressMultiLine id={id} {...addressProps} />
-
+        <Typography variant='body1' component='div'>{addressProps.telephone}</Typography>
```
paales, line 34: Shouldn't we just add it to the AddressMultiline?

mikekeehnen: We could, but this allowed it to be more scoped. AddressMultiline is used quite often throughout the project. Do you think we should add it to AddressMultiline?

PR #2155, opened 2024-01-03
Wrapped RecentlyViewedProducts in AddProductsToCartForm

Jessevdpoel:


Jessevdpoel:
Wrapped the recently viewed products in an AddProductsToCartForm because the RVP uses the ProductWishlistIconButton which uses the AddProductsToCartForm to get the selected options. 

Since there are never any selected options when you add an item to your wishlist from a listproduct maybe we should create two different productwishlistIconButtons? Maybe one for the PDP and one for the listItems? We want to do this only to prevent wrapping the AddProductsToCartForm around the recently viewed products.

examples/magento-graphcms/pages/p/[url].tsx
```diff
@@ -187,7 +187,6 @@ function ProductPage(props: Props) {
           }}
         />
       )}
-
```
Comment (bramvanderholst) [inline examples/magento-graphcms/pages/p/[url].tsx:190]: Please revert this unnecessary change in the examples folder.

bramvanderholst (approved):

PR #2158, opened 2024-01-04
Refactor GA4 / GTM tracking preferences (GCOM-1312)

mikekeehnen:


Giovanni-Schroevers:
@mikekeehnen Could you make all the fragments used in analytics injectable?

packages/analytics/events/index.ts
```diff
@@ -0,0 +1 @@
+
```
paales, line 1: Empty file

paales (changes requested):

PR #2165, opened 2024-01-10
[GCOM-1315] fix cli tool + fix migration 6to7

Jessevdpoel:


paales:
- Users can't rerun 6to7, please add it to 7to8 (should be available after rebasing from canary) (or to 8to9 if this doesn't make the release).
- Remove the fallback for the management url.

paales (changes requested):
- Users can't rerun 6to7, please add it to 7to8 (should be available after rebasing from canary) (or to 8to9 if this doesn't make the release).
- Remove the fallback for the management url.

PR #2167, opened 2024-01-15
Magento Varnish cache purge handling

hnsr:


paales:
@hnsr I've changed this to draft, no idea what to do with this in the current state.

PR #2169, opened 2024-01-16
Fix/gcom 1315 cli migration

Jessevdpoel:


packages/hygraph-cli/src/migrations/graphcommerce6to7.ts
```diff
@@ -303,5 +303,20 @@ export const graphcommerce6to7 = async (schema: Schema) => {
...
+      displayName: 'Conditions',
+      apiId: 'conditions',
+      parentApiId: 'ConditionOr',
+      componentApiIds: ['ConditionText', 'ConditionNumber'],
+      isList: true,
+    },
+    'ConditionOr',
+    'component',
+  )
+
```
paales, line 320: Please remove the migrations from older versions, this version has already been released and should not be modified

.changeset/twelve-socks-tap.md
```diff
@@ -0,0 +1,6 @@
+---
+"@graphcommerce/next-config": patch
+"@graphcommerce/hygraph-cli": patch
+---
+
+Fix hygraph 6to7 cli
```
paales, line 6: Please update the description

paales (changes requested):

PR #2170, opened 2024-01-17
fix(GCOM-1317): fix bug where inactive cached cart could be set as currentCartId

FrankHarland:


.changeset/nasty-icons-knock.md
```diff
@@ -0,0 +1,6 @@
+---
+"@graphcommerce/magento-cart": patch
+"@graphcommerce/magento-customer": patch
+---
+
+fix(GCOM-1317): fix bug where inactive cached cart could be set as currentCartId
```
paales, line 6: Please do not reference internal issues in release notes.

PR #2203, opened 2024-02-13
CheckboxElement, MultiSelectElement, NumberFieldElement, SelectElement, SliderElement and TextFieldElement have their inputRef passed, allowing focus to be set by the form.

Jessevdpoel:
You can test this by adding an item to the cart going to the payment step of the checkout. Make your screen small enough so the checkboxes are not visible and click on pay without checking the checkboxes

### DIFF
--- a/packages/ecommerce-ui/components/FormComponents/CheckboxElement.tsx
+++ b/packages/ecommerce-ui/components/FormComponents/CheckboxElement.tsx
@@ -64,8 +64,8 @@ export function CheckboxElement<TFieldValues extends FieldValues>({
                 label={label || ''}
                 control={
                   <Checkbox
-                    inputRef={checkboxRef}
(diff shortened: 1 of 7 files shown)


packages/ecommerce-ui/components/FormComponents/CheckboxElement.tsx
```diff
@@ -47,7 +47,8 @@ export function CheckboxElement<TFieldValues extends FieldValues>({
       name={name}
       rules={rules}
       control={control}
-      render={({ field: { value, onChange }, fieldState: { invalid, error } }) => {
+      render={({ field: { value, onChange, ref, ...field }, fieldState: { invalid, error } }) => {
```
Jessevdpoel, line 50: ref kan hier weggehaald worden dan

PR #2204, opened 2024-02-13
Feature Add x-forwarded-for header to mesh

SumVur:
Some of the modules use `x-forwarded-for` from order entity to check for fraud


paales:
Can you provide us with a real example? We haven't encountered this requirement. Do you believe every GraphCommerce shop should incorporate this?

paales:
Thanks @SumVur!

PR #2217, opened 2024-02-22
Fix/gcom 1344

JoshuaS98:
Before
image

After
image


packages/magento-product/components/ProductPageBreadcrumb/ProductPageBreadcrumb.tsx
```diff
@@ -20,21 +20,26 @@ export function ProductPageBreadcrumb(props: ProductPageBreadcrumbsProps) {
...
+              </Link>
+            ))}
+
+          <Link href={`/${category?.url_path}`} underline='hover' color='inherit'>
+            {category?.name}
           </Link>
-        ))}
-      <Link href={`/${category?.url_path}`} underline='hover' color='inherit'>
-        {category?.name}
-      </Link>
```
paales, line 37: Only this bit needs to be wrapped in {category && <Link...>}

PR #2222, opened 2024-03-08
Breadcrumbs (GCOM-1347)

LaurensFranken:


paales:
This seems to be a very large bundle size increase, are we doing something wrong here?

LaurensFranken:
> This seems to be a very large bundle size increase, are we doing something wrong here?

We are now importing the Breadcrumbs on Category and Product pages which was not the case before. Maybe this isn't something we want? Then people should know that the package exists and add it to the project itself when they need it

paales:
As discussed, the design should look like this:

Scherm­afbeelding 2024-04-24 om 14 23 21

Also, have it implemented on the Blog?

The header is positioned too high on https://graphcommerce-git-feature-gcom-1347-graphcommerce.vercel.app/nl/heren/fotografie

paales:
Can not close the breadcrumb on mobile

paales:
Superseded by https://github.com/graphcommerce-org/graphcommerce/pull/2273

PR #2223, opened 2024-03-13
Sitemap / robots refactor

bramvanderholst:


examples/magento-graphcms/pages/robots.txt.tsx
```diff
@@ -0,0 +1,91 @@
...
+    `# Sitemaps`,
+    ...(storefront.domain ? 
+      // When the current store has its own domain, only show sitemaps for the specific store
+      sitemaps.map(sitemap => 
+        `Sitemap: ${canonicalize({...options, locale: storefront.locale, defaultLocale: storefront.locale}, `${sitemap}`)}`,
+      ) 
+      : 
+      // When the current store does not have its own domain, show sitemaps for all stores on the same global domain
+      configStorefront.filter(store => !store.domain).map((store) => sitemaps.map(sitemap => 
+        `Sitemap: ${canonicalize({...options, locale: store.locale}, `${sitemap}`)}`,
```
mikekeehnen, line 76: Dit is echt heel moeilijk leesbaar

packages/hygraph-ui/lib/getHygraphPaths.ts
```diff
@@ -0,0 +1,43 @@
...
+          variables: { pageSize, skip: pageSize * (i - 1), where: filter },
+        }),
+      )
+    }
+  }
+
+  const paths: Return['paths'] = (await Promise.all(pages))
+    .map((q) => q.data.pages)
+    .flat(1)
+    .map((page) => ({ params: page, locale }))
```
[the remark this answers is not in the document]
bramvanderholst: As the potential savings in the case are so minimal (probably sub-millisecond), I'd prefer better code readability.

examples/magento-graphcms/pages/robots.txt.tsx
bramvanderholst: I've tried putting it in a single template string, but it did not improve legibility, so unless someone has a cleaner solution, I think I'll leave it as is.

examples/magento-graphcms/pages/sitemap/content.xml.tsx
```diff
@@ -0,0 +1,72 @@
...
+    '*/cart',
+    '*/checkout',
+  ]
+
+  // Filter out category pages as they are already added in sitemap/categories.xml
+  excludes.push(
+    ...(await getCategoryStaticPaths(graphqlSsrClient(locale), locale, { limit: false })).map(
+      (path) => `/${typeof path === 'string' ? path : path.params.url.join('/')}`,
+    ),
+  )
```
paales, line 43: As discussed earlier, might it be possible to get rid of these excludes, this will make it much slower.

paales (changes requested):

examples/magento-graphcms/pages/sitemap/content.xml.tsx
bramvanderholst: This query takes around half a second for a fairly large category tree (uncached - cached is obviously a lot faster).
Seeing as this page is intended for bots only and [bots don't seem to care](https://www.lumar.io/blog/best-practice/google-times-out-after-two-minutes-when-crawling-sitemaps/) about a few seconds more or less, I'd suggest keeping this in to prevent duplicate URLs.

PR #2224, opened 2024-03-15
feat(GCOM-1353): scroll to #products on filter application

FrankHarland:


packages/magento-product/components/ProductFiltersPro/ProductFiltersPro.tsx
```diff
@@ -50,11 +51,41 @@ export function ProductFiltersPro(props: FilterFormProviderProps) {
...
+          setTimeout(() => {
+            window.scrollBy(
+              0,
+              (document.querySelector('.LayoutHeader-root')?.getBoundingClientRect().height ?? 0) *
+                -1,
+            )
+          })
+        })
+      }
+    }
```
paales, line 73: This causes a layout jump if it renders inbetween. Investigate if this works: https://stackoverflow.com/questions/72581132/how-does-the-css-property-scroll-margin-top-and-scroll-padding-top-really-wo

paales (changes requested):

PR #2234, opened 2024-04-02
Upgrade input components to no longer use muiRegister, which improves INP scores

FrankHarland:


packages/react-hook-form/src/useFormGqlQuery.tsx
```diff
@@ -19,7 +18,7 @@ export function useFormGqlQuery<Q extends Record<string, unknown>, V extends Fie
   const tuple = useLazyQuery(document, operationOptions)
   const operation = useFormGql({ document, form, tuple, ...options })
   const muiRegister = useFormMuiRegister(form)
```
FrankHarland, line 20: Should we do anything with this?

paales: I've marked them as deprecated

PR #2242, opened 2024-04-16
Feature/improved pagination

FrankHarland:


packages/next-ui/Pagination/PaginationExtended.tsx
```diff
@@ -0,0 +1,81 @@
...
+          },
+        }),
+        ...(Array.isArray(sx) ? sx : [sx]),
+      ]}
+    >
+      <Pagination
+        count={count}
+        defaultPage={page}
+        page={page}
+        siblingCount={0}
```
paales, line 64: As discussed we'd like to have the siblingCount to be 1 at least. Also we'd like to be able to pass props here.

paales (changes requested):

PR #2286, opened 2024-06-03
chore: add disable option to EditAddressForm

carlocarels90:
**Usage:**
```tsx
<EditAddressForm address={address} disabled />
```
| Before | After |
|--------|--------|
| image | image


paales:
We've going with another feature to make the billing address readonly

PR #2291, opened 2024-06-07
fix: add missing translation for adding multiple products to cart

Giovanni-Schroevers:


paales:
Will be fixed with a `<Plural />` component.

PR #2296, opened 2024-06-17
Fixed Place order button not working (in dev mode) for certain payment methods

bramvanderholst:
Place order button would show a spinner, but nothing would happen. No queries would fire. because the method `PaymentMethodPlaceOrderNoop` was undefined, due to a circular dependency

### DIFF
--- a/packages/magento-payment-included/methods.tsx
+++ b/packages/magento-payment-included/methods.tsx
@@ -1,8 +1,6 @@
-import {
-  PaymentMethodOptionsNoop,
-  PaymentMethodPlaceOrderNoop,
-  PaymentModule,
-} from '@graphcommerce/magento-cart-payment-method'
+import { PaymentModule } from '@graphcommerce/magento-cart-payment-method/Api/PaymentMethod'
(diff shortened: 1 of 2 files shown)


packages/magento-payment-included/methods.tsx
```diff
@@ -1,8 +1,6 @@
-import {
-  PaymentMethodOptionsNoop,
-  PaymentMethodPlaceOrderNoop,
-  PaymentModule,
-} from '@graphcommerce/magento-cart-payment-method'
+import { PaymentModule } from '@graphcommerce/magento-cart-payment-method/Api/PaymentMethod'
+import { PaymentMethodOptionsNoop } from '@graphcommerce/magento-cart-payment-method/PaymentMethodOptionsNoop/PaymentMethodOptionsNoop'
+import { PaymentMethodPlaceOrderNoop } from '@graphcommerce/magento-cart-payment-method/PaymentMethodPlaceOrderNoop/PaymentMethodPlaceOrderNoop'
```
paales, line 3: I'm wondering if this also works:

```suggestion
import {
  PaymentMethodOptionsNoop,
  PaymentMethodPlaceOrderNoop,
  type PaymentModule,
} from '@graphcommerce/magento-cart-payment-method'
```

paales (changes requested):

PR #2302, opened 2024-06-21
Fixed tier prices not working for non-configurable products

bramvanderholst:


ErwinOtten:
For a working demo of pricing tiers: https://graphcommerce.vercel.app/nl/p/babou-gum-gc-1000-sock

Choose size 36-40

PR #2335, opened 2024-07-25
chore: Update the README with clearer instructions for installing MultiSafePay

carlocarels90:


packages/magento-payment-multisafepay/README.md
```diff
@@ -2,20 +2,62 @@
...
+   const { loadConfig } = require('@graphcommerce/next-config')
+
+   const config = loadConfig()
+
+   const nextConfig = {
+     async rewrites() {
+       return [
+         {
+           source: '/multisafepay/connect/:slug*',
+           destination: `${config.magentoBaseUrl}/multisafepay/connect/:slug*`,
```
paales, line 58: https://github.com/graphcommerce-org/graphcommerce/blob/canary/packages/magento-graphql-rest/plugins/meshConfigM2Rest.ts#L14

So something like this?
```
graphCommerceConfig.magentoEndpoint.slice(0, -7)
```

I would like to prevent having to provide the Magento URL multiple times.

paales (changes requested):

packages/magento-payment-multisafepay/README.md
carlocarels90: What do you think of this solution? 

```tsx
config.magentoEndpoint.replace('/graphql', '')
```

machielkreiken: Bump @carlocarels90 @paales

PR #2369, opened 2024-09-12
chore: resolve the issue with active menu items when two or more pages start with the same URL

carlocarels90:


packages/next-ui/LayoutParts/DesktopNavBarItem.tsx
```diff
@@ -42,7 +42,7 @@ export function DesktopNavItem(props: DesktopNavItemLinkProps | DesktopNavItemBu
 
   const { href, children, sx = [], ...linkProps } = props
 
-  const active = router.asPath.startsWith((href ?? '').toString())
+  const active = router.asPath === (href ?? '').toString()
```
paales, line 45: The idea was that if a child is selected that the parent would be selected as well, so that breaks with this, any suggestions?

carlocarels90: @paales see my latest changes 🙂 

Now it activates the parent and only the menu item that exactly matches the `router.asPath`.

https://github.com/graphcommerce-org/graphcommerce/pull/2369/files

PR #2373, opened 2024-09-19
Handle the case where one switches from using the default .html URL suffix, to not using one

hnsr:
When resolving an url `foo.html` inside `redirectOrNotFound`, this change will always add `foo` to the list of candidates to look up, regardless of the current `StoreConfig.product_url_suffix` value.

This allows you to turn off the default `.html` URL suffix in Magento and still have the old URLs redirect correctly.

### DIFF
--- a/.changeset/healthy-lies-act.md
+++ b/.changeset/healthy-lies-act.md
@@ -1,0 +1,5 @@
+---
+'@graphcommerce/magento-store': minor
+---
+
+Handle the case where one switches from using the default .html URL suffix, to not using one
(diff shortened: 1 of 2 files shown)


hnsr:
Tried to check if a custom redirect such as `bar.html` > `men` still worked, but this gives a 404.

Screenshot 2024-09-19 at 09 46 07

This seems to be unrelated to this fix however, also does not seem to work on current demo: https://graphcommerce.vercel.app/en-gb/bar.html

PR #2393, opened 2024-10-23
chore(GCOM-1490): add ‘preventClickBack’ boolean to optionally disable back navigation

carlocarels90:
In some cases, you may want to force the use of the 'up' data for the 'back' button, like the checkout success page. 
This boolean ensures that the 'up' data is always used.

### DIFF
--- a/.changeset/fifty-moose-laugh.md
+++ b/.changeset/fifty-moose-laugh.md
@@ -3,4 +3,4 @@
 '@graphcommerce/next-ui': patch
 ---
 
-Add ‘preventClickBack’ boolean to optionally disable back navigation.
+Solve an issue where the success page would show a back button which would go back to the payment page (which would be empty)
--- a/.changeset/fifty-moose-laugh.md
+++ b/.changeset/fifty-moose-laugh.md
@@ -1,0 +1,6 @@
+---
+'@graphcommerce/magento-graphcms': patch
+'@graphcommerce/next-ui': patch
+---
+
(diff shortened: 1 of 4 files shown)


packages/next-ui/Layout/components/LayoutHeaderBack.tsx
```diff
@@ -12,7 +12,9 @@ import { IconSvg } from '../../IconSvg'
 import { responsiveVal } from '../../Styles'
 import { iconChevronLeft } from '../../icons'
 
-export type BackProps = Omit<LinkOrButtonProps, 'onClick' | 'children'>
+export type BackProps = Omit<LinkOrButtonProps, 'onClick' | 'children'> & {
+  preventClickBack?: boolean
```
paales, line 16: From the code this doesn't really become clear what this does. At least it should have a comment what this does.


```
/**
 * Will not use `router.back()` if available, and will always use the `up.href`
 */
disableBackNavigation?: boolean
```

paales (changes requested):

.changeset/fifty-moose-laugh.md
```diff
@@ -0,0 +1,6 @@
+---
+'@graphcommerce/magento-graphcms': patch
+'@graphcommerce/next-ui': patch
+---
+
+Add ‘preventClickBack’ boolean to optionally disable back navigation.
```
paales, line 6:
```suggestion
Solve an issue where the success page would show a back button which would go back to the payment page (which would be empty)
```

PR #2395, opened 2024-10-25
add disableDeselectOption boolean to prevent users from deselecting configurable options (GCOM-1507)

carlocarels90:
The `disableDeselectOption ` boolean can be used to ensure that users are unable to deselect options within a configurable product. By setting `disableDeselectOption ` to `true`, as in the example below, any selected option will remain selected, preventing users from accidentally or intentionally clearing their choice.

**Usage:**

 ```
<ConfigurableProductOptions product={product} disableDeselectOption />
```

### DIFF
--- a/packages/ecommerce-ui/components/FormComponents/ActionCardListForm.tsx
+++ b/packages/ecommerce-ui/components/FormComponents/ActionCardListForm.tsx
@@ -4,7 +4,7 @@ import React, { MouseEventHandler } from 'react'
 
 export type ActionCardItemBase = Pick<ActionCardProps, 'value'>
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/ecommerce-ui/components/FormComponents/ActionCardListForm.tsx [hunk 1]
```diff
@@ -33,6 +35,7 @@ export function ActionCardListForm<
     multiple,
     disabled,
     shouldUnregister,
+    forceSelection,
```
paales, line 38: I personally like the 'disableXY' naming convention, so something like:`disableDeselectOption`? What do you think?

paales (changes requested):

PR #2406, opened 2024-11-07
Added github actions periodic build

Renzovh:


.github/workflows/periodic-build.yml
```diff
@@ -0,0 +1,96 @@
...
+      slackToken: ${{ secrets.SLACK_BOT_RELEASE_TOKEN }}
+    with:
+      channel: slack-test-messages
+      message: Graphcommerce start building
+      notifyOnlyOnFailure: true
+
+  publish:
+    strategy:
+      matrix:
+        version: [18, 20, 22]
```
paales, line 24: For the main branch, it should be `[18, 20]`

.github/workflows/periodic-build.yml
```diff
@@ -0,0 +1,96 @@
+name: Periodic build
+on:
+  schedule:
+    - cron: 0 0 * * *
+  push:
+    branches:
+      - 'feature/periodic-release'
```
paales, line 7:
```suggestion
      - 'canary'
```

paales (commented):
After we've made this workflow for the main branch and everything works as expected. We create a new PR that targets the canary branch which has the modifications that are specific to that version, specifically node 22.

Issue #2401, opened 2024-11-06
Safari browser version 15.6 The Account page show's client-side exception


vigneshMuthumariyappan:
### Describe the Bug

When the user has the order on his/her account the account page is a break

It happens only in the Safari browser version 15.6

![image](https://github.com/user-attachments/assets/88805849-699b-4f09-a935-b7628a693059)

### Expected Behavior

Page render without client-side exception

### To Reproduce

1 -> Create a user and login
2 -> Add any product to the cart and place an order
3 -> Go to the account page and see the client-side exception

PR #2414, opened 2024-11-12
Fix date browser compatible

Vignesh-M21:
Hi @paales 

This https://github.com/graphcommerce-org/graphcommerce/issues/2401 issue has been fixed here.


paales:
@Vignesh-M21 I've updated the PR with some cleanup, does this look correct?

paales:
@Vignesh-M21 I've added RelativeTimeFormat and RelativeToTimeFormat components to move this toDate logic into the package. Should be good to go :)

PR #2444, opened 2024-12-03
Reduce time of second render in profiler by preventing the use of noSsr / useIsSSR

Giovanni-Schroevers:


paales:
Another PR will be made.

PR #2490, opened 2025-01-24
fix: update label for bank selection to use i18n for localization

carlocarels90:


paales:
@carlocarels90 This becomes irrevant with the issuer being removed

paales:
1

PR #2498, opened 2025-02-07
FIx typo in documentation

TXHENG:
The correct path for Magento Open Source example should be 

```bash
cp -R graphcommerce/examples/magento-open-source/. my-project && cd my-project
```

instead of 

```bash
cp -R graphcommerce/examples/magento-magento-open-source/. my-project && cd my-project
```


paales:
Thanks!

PR #2506, opened 2025-03-14
feat(GCOM-1577): show product name and link on PDP breadcrumbs

FrankHarland:
| Before | After |
|--------|--------|
| Screenshot 2025-03-14 at 14 35 14 | Screenshot 2025-03-14 at 14 38 03 |


paales:
This is done for mobile explicitly as we haven't got much space there. Is this an explicit customer request or something or why would we want this?

PR #2507, opened 2025-03-14
feat(GCOM-1578): fix issue where gtag events would not fire due to faulty plugin loading

FrankHarland:
This was a pretty severe bug in GC 9.0.3
The sendEvent function is properly plugged in the @graphcommerce/google-datalayer plugin, but is not being imported properly because in our components that should fire analytics events we're using useSendEvent() which just loads the default empty function.

Here we're plugging into the useSendEvent (rather than the sendEvent function) to properly fire the events via gtag.

| Before | After |
|--------|--------|
| Screenshot 2025-03-14 at 15 02 01 | Screenshot 2025-03-14 at 15 12 50 |


paales:
@FrankHarland We can also split the sendEvent and useSendEvent into two files, that should fix the issue as well.

PR #2635, opened 2026-05-28
Feature/layout navigation v2

bramvanderholst:
## Summary

Refactors `LayoutNavigation` in the example storefronts from a single ~180-line file
into composable pieces and moves the canonical `LayoutDefault` from
`@graphcommerce/next-ui` into each project's `components/Layout/`. No visual change,
no breaking imports.

## What's in this PR

- **Split** `LayoutNavigation` into `Header/Header.tsx`, `Header/HeaderContainer.tsx`,
  `MenuOverlay.tsx`, and a project-local `LayoutDefault.tsx`. `HeaderContainer` is the
  `<Container component='header'>` wrapper lifted out of `LayoutDefault`, so each project
  fully owns its header container (sticky, mobile pointer-events, styling) without
  forking next-ui.
- **`Header/` and `Footer/`** live in their own subfolders so future additions
  (mega menu, mobile-only header, extra footer columns) don't require further moves.
- **`LayoutMinimal`** uses the same local `LayoutDefault` + `HeaderContainer` as the
  canonical pattern.
- **`disableScrollEffects`** prop added to `CartFab` and `NavigationFab`, which makes
  it easier to drop them into a fully-custom header.
- **`@graphcommerce/next-ui`'s `LayoutDefault`** is marked `@deprecated`. The
  implementation stays in place; the deprecation only points consumers at the local copy
  they already have after a sync.

## Migration

Old monolithic version is kept as `LayoutNavigationLegacy.tsx`. To revert, flip one line
in `components/Layout/index.ts`:

```ts
export * from './LayoutNavigation'
// export * from './LayoutNavigationLegacy'


bramvanderholst:
Superseded by https://github.com/graphcommerce-org/graphcommerce/pull/2637

PR #2655, opened 2026-07-27
PlaceOrderOutput.errors is nullable since Magento 2.4.9

paales:
Magento 2.4.9 / MageOS 3.2.0 relaxed `PlaceOrderOutput.errors` from `[PlaceOrderError]!` to `[PlaceOrderError]` and the resolver now returns `null` when the order is placed successfully (see the schema diff: https://github.com/ho-nl/magento2-graphql-versions/compare/2.4.8...2.4.9#diff-efc1675187620595b0844197a25c35eac9b6df752f9182e5cffddee6f27a8489L8943-R9032).

A mesh whose unified schema still declares the field non-null (introspected from an older backend, or any `magentoVersion` < 249 setup) then fails **every successful order** at the gateway with:

```
Cannot return null for non-nullable field PlaceOrderOutput.errors.
```

while the order *is* placed in Magento — the customer just sees a generic payment error.

### Fix

Add `packages/magento-graphql/schema-249/PlaceOrderOutput.graphqls` redeclaring `errors` as nullable. The `schema-249` additionalTypeDefs are applied for every project with `magentoVersion` < 249, so the unified schema always carries the loosest contract across supported Magento versions:

- 2.4.9 backend + `magentoVersion: 248`: mesh no longer bombs on `errors: null` (this was the bug).
- 2.4.8 backend: still returns a non-null list, which trivially satisfies the nullable field.
- `magentoVersion: 249`: typedef not applied; live introspection is already nullable.

Verified that stitching merges a partial type redefinition as expected: `errors` becomes nullable while `order` / `orderV2` and the rest of the introspected type are preserved. Client code (`assertOrderPlaced`, `PaymentMethodPlaceOrderNoop`) already handles `errors: null`.

### DIFF
--- a/.changeset/place-order-errors-nullable.md
+++ b/.changeset/place-order-errors-nullable.md
@@ -1,0 +1,5 @@
+---
+'@graphcommerce/magento-graphql': patch
+---
+
+Relax `PlaceOrderOutput.errors` to nullable in the unified schema. Magento 2.4.9 / MageOS 3.2.0 changed the field from `[PlaceOrderError]!` to `[PlaceOrderError]` and returns `null` when the order is placed successfully, so a storefront configured with `magentoVersion` < 249 talking to a 2.4.9 backend failed every successful order with `Cannot return null for non-nullable field PlaceOrderOutput.errors.`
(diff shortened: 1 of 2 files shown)


paales:
SLOP

PR #2660, opened 2026-08-06
fix(next-config): redirect Magento account URLs to the GraphCommerce routes

paales:
_Written by Claude Code:_

Magento builds its own frontend URLs from `base_link_url`. On a headless setup that value points at the GraphCommerce storefront, so every stock transactional email ships a link GraphCommerce cannot serve:

```
$ grep -rEho "'[a-z_]+/[a-z_]+/?[a-zA-Z_]*/?'" $(find vendor -path '*/email/*.html')
  48 'customer/account/'
  10 'customer/account/createPassword/'
   2 'customer/account/createPassword'
   2 'customer/account/confirm/'
```

`createPassword` and `confirm` are covered — all three examples ship `pages/customer/account/{confirm,createPassword}` (they are scaffolded from the examples; `@graphcommerce/magento-customer` has no `copy/` directory). `customer/account/` is not, and it is the single most common link in the whole email set ("Sign in to your account", in the footer of every order/invoice/shipment/credit-memo mail). It 404s.

There used to be exactly this redirect, added in #1936 and moved into `withGraphCommerce` shortly after. It was dropped as collateral in "Remove redirects for `/product/$type/[url]` routes" (2c79a4cb) — that commit removed the whole `redirects()` block, and the `/customer/account` entry happened to live in the same array. This restores it and adds the neighbouring URLs that Magento also emits.

## The list

| Magento URL | GraphCommerce route | Where it comes from |
| --- | --- | --- |
| `/customer/account` | `/account` | every stock email template |
| `/customer/account/index` | `/account` | Magento's canonical alias for the dashboard |
| `/customer/account/login` | `/account/signin` | 302 `Location` on gated Magento routes — `magento-product-downloadable`'s copy pages already sniff for this string |
| `/customer/account/create` | `/account/signin` | registration links; `/account/signin` renders both sign-in and sign-up |
| `/sales/order/history` | `/account/orders` | Magento's "My Orders" |

## Why exact matches, not `:path*`

A `redirects()` entry is evaluated **before** the filesystem routes, so redirecting a path that has a page makes that page unreachable. GraphCommerce serves exactly two paths under `/customer/`: `confirm` and `createPassword` — the two carrying Magento's one-shot confirmation `key` and reset `rp_token`. A catch-all would break account confirmation and password reset outright, so every entry stays an exact source. Note that `/customer/account/create` is a distinct route from `/customer/account/createPassword` and does not cover it.

**Dropped after review** (a3a45d2fe9): `/customer/account/forgotpassword` was in the first revision. It shadows nothing — there is no `forgotpassword` page anywhere in the examples — but it appears in zero email templates, and Magento's reset token rides on `createPassword`, not on it. No evidence, so it's gone.

## Notes

- Projects that already define their own `redirects()` keep working: `withGraphCommerce` awaits `nextConfig.redirects()` first and appends, and Next.js takes the first match, so a project override wins.
- `dist/index.js` is updated alongside `src/` per the convention in this package.

### DIFF
--- a/packagesDev/next-config/dist/index.js
+++ b/packagesDev/next-config/dist/index.js
@@ -1660,6 +1660,32 @@ function withGraphCommerce(nextConfig, cwd = process.cwd()) {
         ...nextConfig.images?.remotePatterns ?? []
       ].filter((v) => !!v)
     },
+    redirects: async () => {
+      const redirects = await nextConfig.redirects?.() ?? [];
+      // Magento builds its own frontend URLs from `base_link_url`, which on a
+      // headless setup points at the GraphCommerce storefront. Those URLs end up
+      // in transactional emails ("Sign in to your account" links every stock
+      // email template renders) and in the 302 Location of gated Magento routes,
+      // but GraphCommerce doesn't serve them, so they 404.
+      //
+      // These have to stay exact matches: redirects are evaluated *before* the
+      // filesystem routes, so a `/customer/account/:path*` catch-all would
+      // shadow the pages/customer/account/{confirm,createPassword} routes that
+      // @graphcommerce/magento-customer copies into the project.
+      redirects.push(
+        { source: "/customer/account", destination: "/account", permanent: true },
+        { source: "/customer/account/index", destination: "/account", permanent: true },
+        { source: "/customer/account/login", destination: "/account/signin", permanent: true },
+        { source: "/customer/account/create", destination: "/account/signin", permanent: true },
+        {
+          source: "/customer/account/forgotpassword",
+          destination: "/account/forgot-password",
+          permanent: true
+        },
+        { source: "/sales/order/history", destination: "/account/orders", permanent: true }
+      );
+      return redirects;
+    },
     rewrites: async () => {
       let rewrites = await nextConfig.rewrites?.() ?? [];
       if (Array.isArray(rewrites)) {
--- a/packagesDev/next-config/dist/index.js
+++ b/packagesDev/next-config/dist/index.js
@@ -1668,20 +1668,17 @@ function withGraphCommerce(nextConfig, cwd = process.cwd()) {
       // email template renders) and in the 302 Location of gated Magento routes,
       // but GraphCommerce doesn't serve them, so they 404.
       //
-      // These have to stay exact matches: redirects are evaluated *before* the
-      // filesystem routes, so a `/customer/account/:path*` catch-all would
-      // shadow the pages/customer/account/{confirm,createPassword} routes that
-      // @graphcommerce/magento-customer copies into the project.
+      // A redirect wins over a filesystem route, so every source below must be
+      // a path GraphCommerce does *not* serve. Two Magento paths are real pages
+      // in the examples — /customer/account/confirm and
+      // /customer/account/createPassword, the ones carrying the confirmation
+      // `key` and the reset `rp_token` — so these stay exact matches. A
+      // `/customer/account/:path*` catch-all would make both unreachable.
       redirects.push(
         { source: "/customer/account", destination: "/account", permanent: true },
         { source: "/customer/account/index", destination: "/account", permanent: true },
         { source: "/customer/account/login", destination: "/account/signin", permanent: true },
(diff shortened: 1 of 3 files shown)


packagesDev/next-config/dist/index.js
```diff
@@ -1660,6 +1660,32 @@ function withGraphCommerce(nextConfig, cwd = process.cwd()) {
...
+      redirects.push(
+        { source: "/customer/account", destination: "/account", permanent: true },
+        { source: "/customer/account/index", destination: "/account", permanent: true },
+        { source: "/customer/account/login", destination: "/account/signin", permanent: true },
+        { source: "/customer/account/create", destination: "/account/signin", permanent: true },
+        {
+          source: "/customer/account/forgotpassword",
+          destination: "/account/forgot-password",
+          permanent: true
+        },
```
paales: I think this is incorrect? We already have a path for this in the examples? As the forgotPassword also carries a token etc.

paales: _Written by Claude Code:_

You're right that it should go — dropped in a3a45d2fe9. The reasoning is slightly different from yours though, and the difference matters for the entries I kept, so let me lay out what I actually found.

**There is no `forgotpassword` page.** Full inventory of what GraphCommerce serves under `/customer/`, across all three examples:

```
examples/magento-graphcms/pages/customer/account/confirm.tsx
examples/magento-graphcms/pages/customer/account/createPassword/index.tsx
examples/magento-open-source/pages/customer/account/confirm.tsx
examples/magento-open-source/pages/customer/account/createPassword.tsx
examples/magento-storyblok/pages/customer/account/confirm.tsx
examples/magento-storyblok/pages/customer/account/createPassword.tsx
```

That's it — `confirm` and `createPassword`. `/account/forgot-password` exists, but that's the GraphCommerce route and it was the *destination* of the redirect, not the source. (Also worth correcting something I wrote in the original comment: `@graphcommerce/magento-customer` has no `copy/` directory, so these aren't copy-pages — they're scaffolded from the examples.)

**The token is on `createPassword`, not on `forgotpassword`.** Grepping every stock Magento email template for customer URLs:

```
'customer/account/'                → [_nosid:1]                                   48x
'customer/account/createPassword'  → [_query:[id, token: $customer.rp_token, …]]  12x
'customer/account/confirm/'        → [_query:[id, key: $customer.confirmation, …]] 2x
```

`forgotpassword` appears in zero templates. It's the form where a customer types their email address; the reset link Magento mails out points at `createPassword` with `id` + `rp_token`. So the entry wasn't dangerous — it just had no evidence behind it, which is a good enough reason to cut it.

**Why the rest are safe, stated properly.** The mechanism I should have named up front: a `redirects()` entry is evaluated *before* the filesystem routes, so redirecting a path that has a page makes that page unreachable. `confirm` and `createPassword` are exactly the two pages that carry a one-shot token, so making either unreachable would break account confirmation and password reset outright. That's why every source is an exact match and why a `/customer/account/:path*` catch-all is off the table. Re-audited against the inventory above:

| source | collides with a page? |
| --- | --- |
| `/customer/account` | no |
| `/customer/account/index` | no |
| `/customer/account/login` | no |
| `/customer/account/create` | no — exact match, does not cover `/customer/account/createPassword` |
| `/sales/order/history` | no |

That fourth row is the one that looks alarming at a glance and I suspect is what caught your eye. Next matches `source` against the full path, so `/customer/account/create` and `/customer/account/createPassword` are distinct routes; the redirect fires only on the former. I've made the comment in the code say all of this explicitly instead of hand-waving at "shadowing".

On evidence tiers for what's left, so you can trim further if you'd rather: `/customer/account` is the strong one (48 email templates). `/customer/account/login` is backed in-repo — `magento-product-downloadable`'s copy pages already sniff for that string in Magento's 302 `Location`. `/customer/account/index` is the same controller action as the first, same destination, zero risk. `/customer/account/create` and `/sales/order/history` are the weakest — no direct hit in the mail templates, just canonical Magento entry points people bookmark or land on from an old indexed URL. Happy to drop those two as well if you'd prefer the list to be strictly email-evidenced.

Mirrored downstream in `ho-nl/project-jumbosports-gc` (b48a202) so the project list stays identical to this one.

PR #2662, opened 2026-08-06
fix(storyblok-ui): cut redundant Storyblok CDN requests

paales:
_Written by Claude Code:_

Five changes across `@graphcommerce/storyblok-ui` and `@graphcommerce/graphql`, all aimed at the number of requests a storefront sends to Storyblok. Requests are billed per call against the space quota, so this is a cost problem as much as a performance one.

## 1. The cache-version refresh never refreshed anything

`refreshStoryblokCacheVersion()` requests `cdn/spaces/me` and relies on `storyblok-js-client` to pick the returned version up. That client only advances its pinned `cv` from a response with a **top-level** `cv` field:

```js
if (params.token && response.data.cv) { … cacheVersions[params.token] = response.data.cv }
```

but `cdn/spaces/me` answers with `{ "space": { "version": 1785830443 } }` — no `cv`. Verified against a live space: a fresh client that only calls `cdn/spaces/me` still reports `cacheVersion() === undefined`. The pin stayed at whatever the first `cdn/stories` response set, which is exactly the staleness the function's docblock says it prevents, and the periodic request bought nothing. `space.version` is the same number `cdn/stories` returns as `cv`, so the fix is to read and apply it.

## 2. Unknown slugs cost a request each

@paales asked whether a completely random URL creates an API call, since bots generate them by the thousand. It did, and the per-slug negative cache in the first revision did not help: every made-up URL is a fresh slug, so it never got a second chance.

`fetchStory()` now checks a slug index first. One `cdn/links` request per cache-version lists every published slug in the space (~300 bytes per story, 1000 per request); anything outside that set is reported as not found without a request. Measured against a live space, counting outgoing HTTP requests with the client response cache disabled:

| Scenario | `cdn/stories` | `cdn/links` |
|---|---|---|
| Cold start, 1 known slug | 1 | 1 |
| **300 random bot URLs** | **0** | **0** |
| Known slug | 1 | 0 |
| After the cache-version advances, 1st random URL | 0 | 1 (index rebuilt) |
| 50 more random URLs, same version | 0 | 0 |

Things that could silently break pages, checked:

- **Language scoping is a trap.** `cdn/links?language=de` returned 25 of 63 entries on a space that does not define `de`, while `cdn/stories/<slug>?language=de` resolves all 63 by falling back to the default language. A language-scoped index would report existing pages as missing, so the index is deliberately unscoped.
- **Folders** are skipped. A folder with a start page contributes that start page as its own entry (`clubkleding/`), which keeps `fetchStory('clubkleding')` working, while `global` and `modal` — folders without one — correctly report as missing (both verified 404 against the CDN).
- **It never claims absence when it does not know.** A failed `cdn/links` request, or a space above 10 000 stories, yields "unknown" and the lookup falls through to the CDN. Verified with the index forced unavailable: a known slug still resolves.
- Draft/preview reads and development skip the index entirely.

## 3. Freshness is pushed, not polled

`requestStoryblokCacheVersionRefresh()` — what a cache-notify webhook calls on publish — only flipped a flag in the process that received the webhook. Every other server learned about the change through `cacheVersionTtl`: one `cdn/spaces/me` per server per interval, forever, whether or not anything was published. At the old 60 second default that is ~446 000 billed requests a month on a ten-replica deployment, spent discovering that nothing happened.

The webhook now publishes a renew signal that every server reads before a published read (throttled to once a second, never an API request). A server seeing a newer signal fetches the cache-version once. `cacheVersionTtl` moves to 3600 as the failsafe, which brings the same deployment to roughly 8 000 requests a month while making publishes visible faster than before.

**How the signal travels — and what it leans on.** It goes through `globalThis.__incrementalCache`, so it reaches every server sharing a `cacheHandler`. That is the same requirement a deployment already meets to serve consistent ISR, so there is nothing extra to configure — and no config key, which is the point: an earlier revision of this PR used a file in a `cacheVersionSignalDir`, which only works if every replica mounts the same volume.

Being explicit about the risk, since this is two internal APIs:

- `globalThis.__incrementalCache` is set in `handleRequestImpl` (`next/dist/server/base-server.js`) on every incoming request, ahead of API routes and ISR regeneration, under the same name and shape in Next 13, 14, 15 and 16. Middleware and the edge runtime construct their own instance and are not covered.
- The entry is written as `kind: 'FETCH'` with `fetchCache: true` — the one path that stores a key verbatim instead of running it through `normalizePagePath()`, which is what allows a namespaced key.
- **If either changes, this fails closed**: no global or no cache handler means `set` is a no-op and `get` returns `null`, never a throw, and freshness falls back to `cacheVersionTtl`. Both calls are wrapped anyway, because a non-FETCH value on the key would raise `InvariantError E653`; the key is namespaced so nothing else can write it.
- **The call shape does drift** — `revalidate` became `cacheControl` between Next 14 and 15 — so it needs re-testing on each Next major.
- Deployments on Next's default `FileSystemCache` need `cacheMaxMemorySize: 0`; its in-memory LRU sits in front of the shared layer and would answer every read from the writing process' own memory. Custom cache handlers have no such layer. Documented on `cacheVersionTtl` in `Config.graphqls`.

Cross-pod behaviour validated on a live two-pod cluster: pod A writes, pod B reads the value within 9 ms, and updates propagate on the next read.

## 4. One signal, not two

`graphqlSsrClient()` in the examples solved the same problem its own way: it dropped its per-locale Apollo client when the timestamp in `./tmp/renew-all-pages-query.txt` moved. That path resolves *inside each container*, so on a multi-replica deployment it silently invalidated nothing — the same assumption that made a file wrong for Storyblok. Both now read one signal, so a single publish invalidates both caches.

The helper lives in `@graphcommerce/graphql`: `storyblok-ui` peer-depends on it, it does not depend on `storyblok-ui`, and all three examples already import from it. The key is neutral (`gc:signal:content-renew`) because the fact being published is "content changed".

**The sync/async problem.** `graphqlSsrClient()` is synchronous and called as `const client = graphqlSsrClient(context)` throughout `getStaticProps`; making it async would break every consumer in every project. So `renewSignal()` returns the last read value and schedules the next refresh — staleness is bounded at one poll interval plus one call, which is well inside what this is for.

**Cold start is handled explicitly.** `renewSignal()` returns `undefined` until the first read completes, which is deliberately distinct from "read, and nothing was published". Every pod is in that state right after a deploy — exactly when all pods are replaced at once — and treating it as `0` would silently skip the first invalidation. Callers create the client without invalidating (a client made now cannot predate a publish) and let the scheduled read decide the next call. Concurrent callers share one in-flight read, so a fresh pod under simultaneous load does not open one per `getStaticProps`.

Verified against a stand-in incremental cache: 25 concurrent refreshes cost one read, a second process picks up a publish, and the client is recreated once per publish and stable in between.

## 5. `getStoryblokStaticPaths()` downloaded every story body

It only reads `full_slug`. The content sitemap in the examples calls it from `getServerSideProps`, so every crawler hit re-downloaded the entire space: **412 KB → 62 KB** on the space I measured.

## Compatibility

No signature changes for callers, including `graphqlSsrClient()`. `storyblok.cacheVersionTtl` changes default from 60 to 3600; projects that set it explicitly are unaffected. The slug index, the version pin and the last-known signal are process-local.


paales (commented):
Does this prevent random URL's generating API calls. A completely random URL should NOT create an API call because bots will generate tons of random URL's?

paales:
_Written by Claude Code:_

Good catch, and the honest answer for the version you reviewed is **no**. The negative cache was per-slug, so it only helped when the *same* made-up URL came back. A bot walking random URLs produces a fresh slug every time, which meant a fresh CDN request every time — exactly the case you were asking about, and the one it did least for.

Fixed in ac2dc8e608 by replacing it with a slug index instead of trying to remember misses.

## Mechanism

One `cdn/links` request per cache-version lists every published slug in the space (`~300 bytes` per story, up to 1000 per request). `fetchStory()` checks membership in that set first; anything outside it returns "not found" with **zero** requests. The index is keyed on the pinned cache-version, so publishing anything drops it and it is rebuilt on the next read — a newly published story is resolvable immediately, not after a TTL.

The per-slug negative cache is gone. With the index it only ever covered cases the index already answers for free, so it was one staleness mechanism too many.

## Measured against a live space (63 links, 55 stories)

Counting outgoing HTTP requests, client response cache disabled so nothing is hidden:

| Scenario | `cdn/stories` | `cdn/links` |
|---|---|---|
| Cold start, 1 known slug | 1 | 1 |
| **300 random bot URLs** | **0** | **0** |
| Known slug (`algemene-voorwaarden`) | 1 | 0 |
| After the cache-version advances, 1st random URL | 0 | 1 (index rebuilt) |
| 50 more random URLs, same version | 0 | 0 |

Before this change those 300 random URLs cost 300 `cdn/stories` requests.

## Things I checked because they could silently break pages

- **Language scoping is a trap.** `cdn/links?language=de` returned 25 of 63 entries on a space that does not even define `de`, while `cdn/stories/<slug>?language=de` resolves all 63 by falling back to the default language. A language-scoped index would therefore report existing pages as missing. The index is deliberately unscoped; story *existence* is language-independent, and folder-level translations are separate stories that appear under their own slug.
- **Folders.** Entries with `is_folder: true` are skipped. A folder with a start page contributes that start page as its own non-folder entry (`clubkleding/`), which is what keeps `fetchStory('clubkleding')` working — while `global` and `modal`, folders without one, correctly report as missing (both verified 404 against the CDN).
- **Trailing slashes.** `foo`, `foo/` and `/foo` all normalise to the same key.
- **It never claims absence when it does not know.** A failed `cdn/links` request, or a space above 10 000 stories, yields "unknown" and the lookup falls through to the CDN exactly as before. Verified: with the index forced unavailable, a known slug still resolves.
- **Memoized in module scope, not in the client's response cache** — `apiOptions.cache` is opt-in and off by default, so relying on it would have meant re-fetching the index on every single `fetchStory`.
- Draft/preview reads and `NODE_ENV=development` skip the index entirely, so the Visual Editor sees an unpublished story the moment it is created.

examples/magento-graphcms/lib/graphql/graphqlSsrClient.ts
```diff
@@ -63,24 +63,15 @@ export function graphqlSsrClient(context: GetStaticPropsContext) {
 
   if (context.preview || context.draftMode) return client(context, 'no-cache')
 
-  const dir = './tmp'
+  // `undefined` means the signal has not been read yet, which is the state of every pod right after
+  // a deploy. Creating the client without invalidating is correct there — a client made now cannot
+  // predate a publish — and `renewSignal()` schedules the read that makes the next call decisive.
```
paales: Remove comment please.

examples/magento-open-source/lib/graphql/graphqlSsrClient.ts
```diff
@@ -53,14 +54,24 @@ export function graphqlSharedClient(context: GetStaticPropsContext) {
...
   const locale = context.locale ?? storefrontConfigDefault().locale
   i18nSsrLoader(locale)
 
   if (context.preview || context.draftMode) return client(context, 'no-cache')
-  if (!ssrClient[locale]) ssrClient[locale] = client(context, 'no-cache')
-  return ssrClient[locale]
+
+  // `undefined` means the signal has not been read yet, which is the state of every pod right after
+  // a deploy. Creating the client without invalidating is correct there — a client made now cannot
+  // predate a publish — and `renewSignal()` schedules the read that makes the next call decisive.
```
paales: Remove comment

examples/magento-storyblok/lib/graphql/graphqlSsrClient.ts
```diff
@@ -53,14 +54,24 @@ export function graphqlSharedClient(context: GetStaticPropsContext) {
...
   const locale = context.locale ?? storefrontConfigDefault().locale
   i18nSsrLoader(locale)
 
   if (context.preview || context.draftMode) return client(context, 'no-cache')
-  if (!ssrClient[locale]) ssrClient[locale] = client(context, 'no-cache')
-  return ssrClient[locale]
+
+  // `undefined` means the signal has not been read yet, which is the state of every pod right after
+  // a deploy. Creating the client without invalidating is correct there — a client made now cannot
+  // predate a publish — and `renewSignal()` schedules the read that makes the next call decisive.
```
paales: Remove comment

packages/storyblok-ui/lib/fetch.ts
```diff
@@ -232,6 +257,99 @@ export async function fetchAllStories(params: FetchStoriesParams): Promise<Story
   }
 }
 
+/**
+ * Above this many links the slug index is not built and `fetchStory` falls back to asking the CDN
+ * per slug. `cdn/links` returns at most 1000 entries per request, so this also caps the index build
+ * at ten requests and the retained set at a few hundred kilobytes.
+ */
+const MAX_INDEXED_LINKS = 10000
```
paales: Remove the limit completely, always cache everything, we'll tackle this when we get here. This would explode if this limit is ever exceeded, more RAM is a better in this case.

examples/magento-graphcms/lib/graphql/graphqlSsrClient.ts
paales: _Written by Claude Code:_ Verwijderd in 013ac97025.

examples/magento-open-source/lib/graphql/graphqlSsrClient.ts
paales: _Written by Claude Code:_ Verwijderd in 013ac97025.

examples/magento-storyblok/lib/graphql/graphqlSsrClient.ts
paales: _Written by Claude Code:_ Verwijderd in 013ac97025.

packages/storyblok-ui/lib/fetch.ts
paales: _Written by Claude Code:_ Limiet volledig weg in 013ac97025 — `MAX_INDEXED_LINKS` en de bijbehorende bail zijn eruit, `fetchSlugIndex()` pagineert nu altijd door tot `total`. De index valt alleen nog terug op `null` als de request faalt.

PR #2663, opened 2026-08-06
fix(magento-customer): honor Magento's required-status for the address telephone

paales:
_Written by Claude Code:_

## The problem

`CartAddressInput.telephone` is a non-nullable `String!` in Magento's schema, so the address forms always have to send *something*. The mutation variables however declare it as an optional `String` with a placeholder default:

```graphql
mutation SetShippingBillingAddress($telephone: String = "000 - 000 0000", ...)
```

`handlerFactory` derives `required` from `variable.type.kind === NON_NULL_TYPE`, so `required.telephone` resolves to `false`, the field renders as optional, and `onBeforeSubmit` fills the gap:

```ts
telephone: variables.telephone || '000 - 000 0000',
```

Every checkout where the customer skipped the field therefore stores a fake phone number on the order — it shows up in the Magento admin, in order confirmation mails and in any address template that renders `telephone`.

## Magento already tells us

`attributesForm(formCode: "customer_address_edit")` returns `is_required` per address attribute, and it reflects the `customer/address/telephone_show` store configuration (required / optional / hidden). Against a stock Magento 2.4.9:

```
firstname   | is_required = true
lastname    | is_required = true
company     | is_required = false
street      | is_required = true
city        | is_required = true
country_id  | is_required = true
postcode    | is_required = false
telephone   | is_required = true
vat_id      | is_required = false
```

GraphCommerce already queries this — `useAttributesForm` in `@graphcommerce/magento-store` — but only for the customer account create/edit forms. The address forms never looked at it.

## The change

`ShippingAddressForm`, `EditBillingAddressForm` and `EditAddressForm` each call the hook directly:

```ts
const addressAttributes = useAttributesForm({ formCode: 'customer_address_edit' })
const telephoneRequired = addressAttributes.find((a) => a.code === 'telephone')?.is_required
```

- The telephone field is marked `required` when Magento says it is required.
- **The placeholder is gone.** An empty telephone is submitted as an empty string, which satisfies the non-nullable input and is validated by Magento against the very same flag: `Magento\Customer\Model\Address\Validator\General::checkOptionalFields()` checks `eavConfig->getAttribute('customer_address', 'telephone')->getIsRequired()`, and `customer/address/telephone_show` is what writes that flag (`Show\AddressOnly`: `req` → 1, `opt` → 0, hidden → 0).
- The stored placeholder is stripped from the form default values in all three forms via a local `legacyPlaceholderTelephone` constant, so a customer editing an address that still carries it gets an empty required field rather than a prefilled fake number. Nothing writes that string anymore.

### Behavior change

Where a customer who bypassed the client-side validation previously ended up with a *successful* order carrying `000 - 000 0000`, a shop that requires a telephone now returns a proper validation error. Verified against Magento 2.4.9 with `telephone_show = req`:

| `telephone` sent to `setShippingAddressesOnCart` | result |
| --- | --- |
| `""` | `"telephone" is required. Enter and try again.` |
| `"0612345678"` | accepted |

That window is small — the preload below puts the metadata in the Apollo cache before first render — and failing loudly beats silently storing junk.

### Version gate

`attributesForm` only exists since Magento 2.4.7, so all of the above is gated in one helper, `packages/magento-customer/utils/legacyPlaceholderTelephone.ts`:

- `stripLegacyPlaceholderTelephone(telephone)` — clears a stored placeholder when an address is loaded into a form; returns the value untouched below 2.4.7.
- `applyLegacyPlaceholderTelephone(telephone)` — submits `''` for an empty field; falls back to the placeholder below 2.4.7.

The required-status derivation needs no extra gate: below 2.4.7 `attributesForm` yields nothing, so `telephoneRequired` stays `undefined` and the field renders exactly as before. Net effect on < 2.4.7 is identical to today.

`CreateCustomerAddressForm` needs no change — its mutation declares `$telephone: String!`, so it was already required.

## Preloading

The metadata is preloaded in `getStaticProps`, the same way `account/signin.tsx` and `account/name/index.tsx` already preload `customer_account_create` / `customer_account_edit`:

```ts
if (magentoVersion >= 247) await preloadAttributesForm(client, 'customer_address_edit')
```

Applied to `checkout/index.tsx`, `checkout/edit/billing-address.tsx`, `checkout/customer/addresses/edit.tsx` and `account/addresses/edit.tsx` in all three examples. Besides saving a client-side round trip, this closes the window where a customer could submit the form before the metadata resolved and still end up with the placeholder.

## Notes / deliberately out of scope

- The `$telephone: String = "000 - 000 0000"` defaults in `SetShippingAddress.graphql`, `SetShippingBillingAddress.graphql` and `SetBillingAddress.graphql` are left as-is. They are unreachable through the forms (`useFormGql` always sends the variable) and removing the default would break any consumer that calls the mutation directly without a telephone.
- Only `telephone` is wired up. The same metadata can drive `company` / `vat_id` / `postcode` required-status later, but `postcode` in particular is country-dependent in Magento (`is_required = false` while the actual requirement comes from the zip-optional country list), so that deserves its own change.


packages/magento-customer/hooks/useAddressAttributes.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/**
+ * Whether Magento requires a telephone number on an address, as configured by
+ * `customer/address/telephone_show`. Returns `undefined` while the attribute metadata is still
+ * loading.
+ */
+export function useTelephoneRequired(): boolean | undefined {
+  return useAddressAttributesRequired().telephone
+}
```
paales: No don't do this. Can you just directly use onst attributes = useAttributesForm({ formCode: addressAttributesFormCode }) in that form to fetch the information. Make sure the information is preloaded as well via getStaticProps like is done for customers.

packages/magento-customer/components/EditAddressForm/EditAddressForm.tsx
```diff
@@ -92,7 +97,7 @@ export function EditAddressForm(props: EditAddressFormProps) {
         <FormRow>
           <TelephoneElement
             variant='outlined'
-            required={required.telephone}
+            required={required.telephone || telephoneRequired === true}
```
paales, line 101: So use the attributes hook more directly.

packages/magento-customer/hooks/useAddressAttributes.ts
paales: _Written by Claude Code:_

Done in 4b2415a2f0741de57ae3158c3e594ecdbcd4e57d — `hooks/useAddressAttributes.ts` is deleted entirely (`useAddressAttributesRequired`, `useTelephoneRequired` and the `placeholderTelephone` constant). Each form now does it itself:

```ts
const addressAttributes = useAttributesForm({ formCode: 'customer_address_edit' })
const telephoneRequired = addressAttributes.find((a) => a.code === 'telephone')?.is_required
```

in `ShippingAddressForm`, `EditBillingAddressForm` and `EditAddressForm`. The placeholder string is inline again, the way the surrounding code already had it.

Preloading follows the `customer_account_create` / `customer_account_edit` pattern from `account/signin.tsx` and `account/name/index.tsx`:

```ts
if (magentoVersion >= 247) await preloadAttributesForm(client, 'customer_address_edit')
```

added to the `getStaticProps` of `checkout/index.tsx`, `checkout/edit/billing-address.tsx`, `checkout/customer/addresses/edit.tsx` and `account/addresses/edit.tsx` in all three examples (12 pages).

That also closes the open point from my previous reply: without the preload there was a window where a customer could submit the form before the attribute metadata resolved, and still get the placeholder. With the metadata in the Apollo state on first render, that window is gone.

packages/magento-customer/components/EditAddressForm/EditAddressForm.tsx
paales: _Written by Claude Code:_

Done in 4b2415a2f0741de57ae3158c3e594ecdbcd4e57d — `EditAddressForm` now calls `useAttributesForm({ formCode: 'customer_address_edit' })` directly and reads `is_required` off the telephone attribute, no wrapper hook in between. Same in `ShippingAddressForm` and `EditBillingAddressForm`, and `account/addresses/edit.tsx` + `checkout/customer/addresses/edit.tsx` preload the metadata in `getStaticProps`.

packages/magento-cart/components/EditBillingAddressForm/EditBillingAddressForm.tsx
```diff
@@ -36,7 +41,7 @@ export function EditBillingAddressForm(props: EditBillingAddressFormProps) {
       city: address?.city,
       countryCode: address?.country.code,
       street: address?.street?.[0] ?? '',
-      telephone: address?.telephone,
+      telephone: address?.telephone !== '000 - 000 0000' ? address?.telephone : '',
```
paales: So the GraphQL query needs a string, but Magneto will now respond correctly right. We did a fallback because we couldn't make the value conditionally required right, so the `000 - 000 0000` isn't necessary anymore and we are just stripping it right now?

paales: _Written by Claude Code:_

Exactly right, and it's gone now — 3ed64083c6f2ab739cbf33a5b4c3025d545f0d63.

All three forms submit `variables.telephone || ''`. The empty string satisfies the non-nullable `CartAddressInput.telephone`, and Magento validates it against the *same* flag we render the field with:

```php
// Magento\Customer\Model\Address\Validator\General::checkOptionalFields()
if ($this->isTelephoneRequired() && !ValidatorChain::is($address->getTelephone(), NotEmpty::class)) {
    $errors[] = __('"%fieldName" is required. Enter and try again.', ['fieldName' => 'telephone']);
}
// isTelephoneRequired() === eavConfig->getAttribute('customer_address', 'telephone')->getIsRequired()
```

and `customer/address/telephone_show` writes into that very attribute (`Magento\Customer\Model\Config\Backend\Show\AddressOnly`): `req` → `is_required = 1`, `opt` / `1` → `is_required = 0`, empty (hidden) → `is_required = 0`. So `attributesForm`'s `is_required`, the required field in the form, and the server-side validator are one and the same source of truth.

**Verified** against Magento 2.4.9 with `telephone_show = req`, `setShippingAddressesOnCart` on a guest cart:

| `telephone` | result |
| --- | --- |
| `""` | `"telephone" is required. Enter and try again.` |
| `"0612345678"` | cart updated, `telephone: "0612345678"` |

With `telephone_show = opt` the branch above is skipped entirely, so `""` is simply accepted — I couldn't flip the config on that environment, hence the code reference rather than a second table row.

**The behavior change is deliberate and worth calling out:** where the old code turned a bypassed client-side validation into a *successful* order carrying junk data, the new code returns a real validation error. That window is small — `preloadAttributesForm` in `getStaticProps` means the metadata is in the Apollo cache on first render, so `required` is set before the customer can type — but if it ever opens, failing loudly beats silently storing `000 - 000 0000`.

**What is still stripped, and why:** each form keeps a local

```ts
const legacyPlaceholderTelephone = '000 - 000 0000'
```

used only when *reading* an address into the form. Addresses saved before this change still carry the placeholder in Magento; showing it prefilled would let a customer submit it again as if it were their number. Clearing it means they get an empty required field. It's read-only legacy-data handling — nothing writes that string anymore.

One thing I left alone: the `$telephone: String = "000 - 000 0000"` defaults still sit in `SetShippingAddress.graphql`, `SetShippingBillingAddress.graphql` and `SetBillingAddress.graphql`. They're unreachable through the forms because `useFormGql` always sends the variable, but if you want the string gone from the package entirely I'm happy to drop those defaults too — it does mean a consumer calling the mutation directly without a telephone starts getting a null-into-`String!` error instead of a silent placeholder.

paales: _Written by Claude Code:_

Fair point — gated now, in one helper: 036e1295a5bbca5b051b21cef89b056cce9041b1.

`packages/magento-customer/utils/legacyPlaceholderTelephone.ts` holds both directions and the version check (`magentoVersion >= 247` from `@graphcommerce/next-config/config`, same import the package already uses in `SignUpForm` and `CompanyFields`):

- `stripLegacyPlaceholderTelephone(telephone)` — clears a stored `000 - 000 0000` when an address is loaded into a form; **returns the value untouched below 2.4.7**, where that string is still actively being written.
- `applyLegacyPlaceholderTelephone(telephone)` — submits `''` for an empty field so Magento validates it; **falls back to the placeholder below 2.4.7**, where the form can never know the field is required and an empty string would just start failing the mutation.

`ShippingAddressForm`, `EditBillingAddressForm` and `EditAddressForm` import those two instead of each carrying a local constant.

The required-status derivation needs no extra gate: below 2.4.7 `attributesForm` yields nothing, so `addressAttributes.find((a) => a.code === 'telephone')?.is_required` stays `undefined`, `required={required.telephone || telephoneRequired === true}` stays `false`, and the field renders exactly as it did before. I kept the `useAttributesForm` call itself ungated to match `useCustomerCreateForm`, which also calls it unconditionally (the `AttributesForm` document already ships in `magento-store` regardless of version) — say the word if you'd rather have a `skip` option on the hook and I'll add it.

So on < 2.4.7 the net behavior is identical to before this PR; on >= 2.4.7 the placeholder is gone from the write path entirely.

PR #2668, opened 2026-08-10
fix(date): parse d/m/Y format and avoid mangling ISO offset strings in toDate

hsngdz:
## What
Fixes two parsing bugs in `toDate`.

## Why
- ISO datetime strings with a timezone offset (e.g. `2024-01-15T10:30:00-05:00`) were silently corrupted into `Invalid Date`. The original code ran `replace(/-/g, '/')` on every string, not just plain `YYYY-MM-DD` dates, so the offset's hyphen got mangled too.
- Strings in Magento's `DATETIME_SLASH_PHP_FORMAT` (`d/m/Y H:i:s`) were misparsed by the native `Date` constructor, which assumes `MM/DD/YYYY` for slash-separated strings. This produced either the wrong date (for ambiguous values like `01/02/2024`) or `undefined` (for unambiguous ones like `15/01/2024`).

## Changes
- Replaced the `replace(/-/g, '/')` string-hack entirely with explicit regex parsing + numeric `Date` construction (`fromComponents`) for both `YYYY-MM-DD` and `d/m/Y H:i:s` formats
- `d/m/Y H:i:s` parsing matches Magento's confirmed day-first format ([`DATETIME_SLASH_PHP_FORMAT`](https://github.com/magento/magento2/blob/2.4.9/lib/internal/Magento/Framework/Stdlib/DateTime.php))
- `fromComponents` also round-trips the constructed date against its inputs, so out-of-range values (e.g. `31/02/2026`) are correctly rejected instead of being silently normalized by JS into a different, valid-looking date

## Test plan
- Added unit tests covering: null/undefined, `Date` instances (valid/invalid), numeric timestamps, `YYYY-MM-DD`, ISO with offset (regression case), `d/m/Y H:i:s` with/without time, ambiguous `d/m/Y` values, invalid strings, and invalid calendar dates

### DIFF
--- a/.changeset/perky-horses-notice.md
+++ b/.changeset/perky-horses-notice.md
@@ -1,0 +1,8 @@
+---
+'@graphcommerce/next-ui': patch
+---
+
+Fix two bugs in `toDate`:
+
+- ISO datetime strings with a timezone offset (e.g. `2024-01-15T10:30:00-05:00`) were being corrupted into `Invalid Date`, because `replace(/-/g, '/')` was applied to every string instead of only to plain `YYYY-MM-DD` dates
+- Strings in Magento's `DATETIME_SLASH_PHP_FORMAT` (`d/m/Y H:i:s`) were misparsed as `MM/DD/YYYY` by the native `Date` constructor, silently producing the wrong date or `undefined`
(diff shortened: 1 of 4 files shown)


paales:
Nice!

